PR #20, opened 2024-06-13
Remaining BouncyCastle translation 

hugoqnc:
**Add most of the missing translation for BouncyCastle assets** (except for a few hard cases that will be handled in a future PR).

In detail:
- Translates Signer and MessageSigner classes
- Translates StreamCipher classes
- Translates Wrapper classes
- Updates some previous translations


engine/src/main/java/com/ibm/engine/utils/DetectionStoreLogger.java
```diff
@@ -162,7 +162,7 @@ private String getDetectionValueContextMessage(
 
     @Nonnull
     String getFormattedNumericString(
-            @Nonnull int hashInt, @Nonnull boolean canBeNegative, @Nullable Integer maxCharacters) {
+            int hashInt, boolean canBeNegative, @Nullable Integer maxCharacters) {
```
n1ckl0sk0rtge, line 165: @hugoqnc is there a reason, why this annotation should be removed?

n1ckl0sk0rtge (approved):
LGTM

engine/src/main/java/com/ibm/engine/utils/DetectionStoreLogger.java
hugoqnc: Primitive data types cannot hold null values anyways, and on VScode with "null analysis" enabled, it is an error (not a warning)!

PR #33, opened 2024-07-03
Update readme

n1ckl0sk0rtge:
- explicitly mentions that only one jar file should be copied
- add note, that later one can install plugin over the marketplace


hugoqnc (approved):
LGTM

PR #36, opened 2024-07-03
Fix Python test failing when upgrading `junit-jupiter-api`

hugoqnc:
Adds the following dependency:
```
        <dependency>
            <groupId>org.junit.platform</groupId>
            <artifactId>junit-platform-launcher</artifactId>
            <version>1.11.0-M2</version>
            <scope>test</scope>
        </dependency>
```
as suggested [here](https://stackoverflow.com/questions/77241793/nosuchmethoderror-java-util-set-org-junit-platform-engine-testdescriptor-getan).

This allows to fix the Python tests when upgrading `junit-jupiter-api` to a version higher than 5.9.3. 
This also fixes the CI problem in "Update dependency graph" which was just a  consequence of the Python tests failing.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -32,4 +32,4 @@ jobs:
 
     # Optional: Uploads the full dependency graph to GitHub to improve the quality of Dependabot alerts this repository can receive
     - name: Update dependency graph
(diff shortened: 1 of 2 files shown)


n1ckl0sk0rtge (approved):
LGTM

PR #68, opened 2024-07-22
Refactoring/general code imporvements

n1ckl0sk0rtge:
- extract sonar issue reporting functionality to higher level rule
- add `SecureRandomGetInstance` rule to java
- add `ITranslationProcess` to python module
- enhance logging
- add java docs
- refactoring


java/src/main/java/com/ibm/plugin/rules/detection/jca/JcaDetectionRules.java
```diff
@@ -54,8 +53,6 @@ public static List<IDetectionRule<Tree>> rules() {
                         // secret key
                         JcaSecretKeyFactoryGetInstance.rules().stream(),
                         JcaSecretKeySpec.rules().stream(),
-                        // prng
-                        JcaSecureRandom.rules().stream(),
```
hugoqnc, line 58: Is it voluntary to have removed this without replacing it with `SecureRandomGetInstance` rules?

n1ckl0sk0rtge: @hugoqnc thanks! For now, yes. This shouldn't be an entry rule. When I refine the java rules, I will but it at the right spot as a depending detection rule.

hugoqnc (approved):
LGTM

PR #76, opened 2024-07-31
Fix/sonar 10.x support

n1ckl0sk0rtge:
Add support for SonarQube 10.x.

I have to heck if this build of the plugin can also be used for sonar 9.x (my assumption is, yes, but I have to test it).

If that works, we can:

- [ ] create a 2.0.0 release
- [ ] remove `dev/1.x.x`


n1ckl0sk0rtge:
this version is backwards compatible with SonarQube 9.x

hugoqnc (approved):
LGTM

PR #131, opened 2024-09-04
Fix/unit tests

n1ckl0sk0rtge:
- remove unnecessary tests
- fix JCA unit tests
- add TestBase class to output module


hugoqnc (approved):
LGTM

PR #134, opened 2024-09-04
Enhance CBOM generation

n1ckl0sk0rtge:
- update output layer
- update naming for algorithms like `AES` and `ChaCha20Poly1305`


hugoqnc (approved):
LGTM

PR #135, opened 2024-09-04
Password Based Encryption model

hugoqnc:
- Adds 3 kinds of PBE: `PBES1`, `PBES2` and `PKCS12PBE`
- I couldn't find which exact PBE is used in JCE, but I mapped it to `PBES1`, which is the only one using the naming `pbeWithXXXAndYYY`

### DIFF
--- a/mapper/src/main/java/com/ibm/mapper/model/PBES1.java
+++ b/mapper/src/main/java/com/ibm/mapper/model/PBES1.java
@@ -36,21 +36,18 @@ public class PBES1 extends Algorithm implements PasswordBasedEncryption {
     @Override
     @Nonnull
     public String asString() {
-        StringBuilder builtName = new StringBuilder(this.name);
-
-        Optional<INode> messageDigest = this.hasChildOfType(MessageDigest.class);
-        Optional<INode> mac = this.hasChildOfType(Mac.class);
-        Optional<INode> cipher = this.hasChildOfType(BlockCipher.class);
+        final Optional<INode> messageDigest = this.hasChildOfType(MessageDigest.class);
+        final Optional<INode> mac = this.hasChildOfType(Mac.class);
+        final Optional<INode> cipher = this.hasChildOfType(BlockCipher.class);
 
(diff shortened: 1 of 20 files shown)


n1ckl0sk0rtge (dismissed):

n1ckl0sk0rtge:
@hugoqnc did some changes, can you check again and review?

mapper/src/main/java/com/ibm/mapper/model/PBES1.java
```diff
@@ -36,21 +36,18 @@ public class PBES1 extends Algorithm implements PasswordBasedEncryption {
...
         } else if (mac.isPresent() && cipher.isPresent()) {
-            return "pbeWith" + mac.get().asString() + "And" + cipher.get().asString();
+            return "PBEWith" + mac.get().asString() + "And" + cipher.get().asString();
         } else if (mac.isPresent()) {
-            return "pbeWith" + mac.get().asString();
+            return "PBEWith" + mac.get().asString();
         }
-
-        return builtName.toString();
+        return this.name;
```
hugoqnc: I used `pbeWith...` rather than `PBEWith...` because of how they write it in the [RFC](https://datatracker.ietf.org/doc/html/rfc2898#appendix-A.3)

n1ckl0sk0rtge: @hugoqnc alright, makes sense. That let's use the notion from the rfc.

mapper/src/main/java/com/ibm/mapper/model/algorithms/PBKDF1.java
```diff
@@ -46,10 +46,14 @@ public String asString() {
...
         }
-
         return this.name;
     }
 
+    @Override
+    public @NotNull String getName() {
+        return asString();
+    }
+
```
hugoqnc, line 56: I'm not sure why you would have to redefine `getName` and cannot only rely on `asString` for this case (and for `BPKDF2`)?

n1ckl0sk0rtge (approved):

PR #137, opened 2024-09-05
Update model

n1ckl0sk0rtge:


mapper/src/main/java/com/ibm/mapper/model/algorithms/AES.java
```diff
@@ -44,24 +44,23 @@ public final class AES extends Algorithm implements BlockCipher, AuthenticatedEn
...
         this.put(new BlockSize(128, detectionLocation));
         this.put(mode);
     }
 
     public AES(int keyLength, @Nonnull Mode mode, @NotNull DetectionLocation detectionLocation) {
-        super(NAME, BlockCipher.class, detectionLocation);
+        this(detectionLocation);
         this.put(new KeyLength(keyLength, detectionLocation));
         this.put(new BlockSize(128, detectionLocation));
         this.put(mode);
```
hugoqnc, line 66: I think the `this.put(new BlockSize(128, detectionLocation));` shouldn't be there, as calling `this(detectionLocation)` will itself call `this(BlockCipher.class, detectionLocation)`, which is already adding the block size

mapper/src/main/java/com/ibm/mapper/model/algorithms/AES.java
```diff
@@ -72,7 +71,7 @@ public AES(
             @Nonnull Mode mode,
             @Nonnull Padding padding,
             @NotNull DetectionLocation detectionLocation) {
-        super(NAME, BlockCipher.class, detectionLocation);
+        this(detectionLocation);
         this.put(new KeyLength(keyLength, detectionLocation));
         this.put(new BlockSize(128, detectionLocation));
```
hugoqnc, line 76: Same here

mapper/src/main/java/com/ibm/mapper/model/algorithms/DH.java
```diff
@@ -44,7 +44,7 @@ public final class DH extends Algorithm implements Signature, KeyAgreement, Publ
     }
 
     public DH(@NotNull DetectionLocation detectionLocation) {
-        super(NAME, PublicKeyEncryption.class, detectionLocation);
+        this(PublicKeyEncryption.class, detectionLocation);
         this.put(new Oid("1.2.840.113549.1.3.1", detectionLocation));
```
hugoqnc, line 48: Same here, no need for the OID

hugoqnc (changes requested):

n1ckl0sk0rtge:
@hugoqnc applied changes, thanks!

hugoqnc (approved):

PR #141, opened 2024-09-11
Fix issue #16

### LINKED ISSUE #16: Next parameter depending rules may not be handled
## The problem
Let's consider this code line example:
```java
GCMBlockCipher blockCipher = GCMBlockCipher.newInstance(AESEngine.newInstance())
```
Suppose we have a rule R1 to detect the call `GCMBlockCipher.newInstance`, and it has a depending detection rule R2 on its parameter to detect the call `AESEngine.newInstance`.
When R1 makes a finding, [this code](https://github.ibm.com/CryptoDiscovery/sonar-cryptography/blob/51d3f15a18ac3cea5903206c5502da2fe2ea568f/engine/src/main/java/com/ibm/engine/language/java/JavaDetectionEngine.java#L746-L759) gets ultimately executed to handle depending detection rules (where `expression` is the argument, so here `AESEngine.newInstance()`):

However, because `AESEngine.newInstance()` is here a `MethodInvocationTree`, it will execute the first case and try to `resolveValuesInOuterScope`, without handling depending rules.

Note that if our code uses a intermediary variable `engine` instead (like below), this problem does not occur and depending rules are correctly handled.
```java
AESEngine engine = AESEngine.newInstance();
GCMBlockCipher blockCipher = GCMBlockCipher.newInstance(engine)
```

## Solution?
A trivial solution idea would be to just handle depending rules in any case, so removing the last line from the `else` statement, giving the following code:
```java
if (expression instanceof MethodInvocationTree methodInvocationTree) {
    // methods are part of the outer scope
    resolveValuesInOuterScope(methodInvocationTree, parameter);
} else if (expression instanceof NewClassTree newClassTree
        && assignedSymbol.isEmpty()) {
    // follow expression directly, do not find matching expression in the method
    // scope
    detectionStore.onDetectedDependingParameter(
            parameter, newClassTree, DetectionStore.Scope.EXPRESSION);
}
// handle next rules
detectionStore.onDetectedDependingParameter(
        parameter, expressionTree, DetectionStore.Scope.ENCLOSED_METHOD);
```
However, while this easy fix correctly fixes our example above, it breaks 3 JCA unit tests (for reasons I have not investigated).

n1ckl0sk0rtge:
@hugoqnc I've checked that this doesn't break any other tests in Java, but I haven't checked it for BC. Please do so before we merge this PR.

Should fix #16

### DIFF
--- a/engine/src/main/java/com/ibm/engine/language/java/JavaDetectionEngine.java
+++ b/engine/src/main/java/com/ibm/engine/language/java/JavaDetectionEngine.java
@@ -19,7 +19,15 @@
  */
 package com.ibm.engine.language.java;
 
-import com.ibm.engine.detection.*;
+import com.ibm.engine.detection.DetectionStore;
+import com.ibm.engine.detection.DetectionStoreWithHook;
+import com.ibm.engine.detection.Handler;
+import com.ibm.engine.detection.IDetectionEngine;
+import com.ibm.engine.detection.MatchContext;
+import com.ibm.engine.detection.MethodDetection;
+import com.ibm.engine.detection.ResolvedValue;
+import com.ibm.engine.detection.TraceSymbol;
+import com.ibm.engine.detection.ValueDetection;
 import com.ibm.engine.hooks.EnumHook;
 import com.ibm.engine.hooks.MethodInvocationHookWithParameterResolvement;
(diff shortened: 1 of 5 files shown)


hugoqnc (approved):
LGTM: it does not break any BC test, and it succeeds on the test `NextParameterDependingRulesTest` which I wrote some time ago specifically to test this issue

PR #142, opened 2024-09-11
Fix issue #12

### LINKED ISSUE #12: Depending detection rules can lead to duplicate findings
Let's look at the example below, where our rule `NEW_CIPHER` (actual rule from Python's `CryptographyCipher.java` file) has two parameters that need to be detected:
```python
private static final IDetectionRule<Tree> NEW_CIPHER = new DetectionRuleBuilder<Tree>()
    .createDetectionRule()
    .forObjectTypes("cryptography.hazmat.primitives.ciphers")
    .forMethods("Cipher")
    .withMethodParameter("cryptography.hazmat.primitives.ciphers.algorithms.*")
        .shouldBeDetectedAs(new AlgorithmFactory())
    .withMethodParameter("cryptography.hazmat.primitives.ciphers.modes.*")
        .shouldBeDetectedAs(new AlgorithmFactory())
    .buildForContext(new CipherContext())
    .withDependingDetectionRules(followingNewCipherRules);
```
This rule has a set of depending detection rules `followingNewCipherRules` at the level of the method (and not related to a parameter).
With the current handling of depending detection rules, once a parameter is detected, the function `analyseExpression` is called, which will call `onReceivingNewDetection`.
Here, both "method parameter related detection rules" and "invoked object related detection rules" are handled using `followNextRules`. 
In our particular case, `onReceivingNewDetection` will be called twice, once for each parameter, which will therefore duplicate "invoked object related detection rules" and their related findings.

Therefore, calling "invoked object related detection rules" should not be done at this point. This may be fixed in the more general refactoring of entry points described in issue #42 or in the more general refactoring of subrules handling described in issue #34.

n1ckl0sk0rtge:
@hugoqnc I've checked that this doesn't break any other tests in Java, but I haven't checked it for BC. Please do so before we merge this PR.

Should fix #12

## How the fix works
If a `shouldBeDetectedAs` defined as part of a detection rule and if this detection should be append as a child to another value (by using `asChildOfParameterWithId(...)`) then this detection value does not get the root depending detection rules appended (done with `withDependingDetectionRules`). 

If more then one detection value should be detected on root level (both don't use `asChildOfParameterWithId(...)`) then, this will still create duplicated nodes on the lower levels. But as part of the translation procedure, this should be merged automatically.

### DIFF
--- a/engine/src/main/java/com/ibm/engine/detection/DetectionStore.java
+++ b/engine/src/main/java/com/ibm/engine/detection/DetectionStore.java
@@ -222,14 +222,25 @@ public class DetectionStore<R, T, S, P> implements IHookDetectionObserver<R, T, 
     }
 
     void addValue(int index, @Nonnull final IValue<T> iValue) {
-        this.detectionValues.compute(
+        addValue(this, index, iValue);
+    }
+
+    void addValue(
+            @Nonnull DetectionStore<R, T, S, P> detectionStore,
+            int index,
+            @Nonnull final IValue<T> iValue) {
+        detectionStore.detectionValues.compute(
                 index,
                 (i, list) -> {
                     if (list == null) {
+                        // If the list is null, create a new ArrayList
+                        // with the iValue and return it
                         final List<IValue<T>> values = new ArrayList<>();
                         values.add(iValue);
                         return values;
                     } else {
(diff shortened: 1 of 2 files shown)


hugoqnc (approved):
LGTM, does not break any BC test

PR #143, opened 2024-09-11
Fix issue #127

### LINKED ISSUE #127: `deepCopy` does not create an instance of the specific class
`Algorithm` contains this `deepCopy` method:
```java
    @Nonnull
    @Override
    public INode deepCopy() {
        Algorithm copy = new Algorithm(this);
        for (INode child : this.children.values()) {
            copy.children.put(child.getKind(), child.deepCopy());
        }
        return copy;
    }
```

All the algorithms (AES, ...) are extending `Algorithm`, so they get this `deepCopy` method. However, calling it on an `AES` object does not create an `AES` instance but an `Algorithm` instance (creating later problems when relying on `instanceof`, in the `enricher` for example).

To solve this, the only solution is probably to add each subclass of `Algorithm` a "copy" constructor (like `AES(AES aes)`, calling the super "copy" constructor of `Algorithm` given below) and to overwrite the  `deepCopy()` method, in which we instantiate a new `AES` using this constructor.

```java
    private Algorithm(@Nonnull Algorithm algorithm) {
        this.children = new HashMap<>();
        this.kind = algorithm.kind;
        this.detectionLocation = algorithm.detectionLocation;
        this.name = algorithm.name;
    }
```

n1ckl0sk0rtge:
- add test case for issue

### DIFF
--- a/mapper/src/main/java/com/ibm/mapper/model/algorithms/RSA.java
+++ b/mapper/src/main/java/com/ibm/mapper/model/algorithms/RSA.java
@@ -57,14 +57,12 @@ public final class RSA extends Algorithm implements KeyAgreement, Signature, Pub
     }
 
     public RSA(@NotNull DetectionLocation detectionLocation) {
-        super(NAME, PublicKeyEncryption.class, detectionLocation);
-        this.put(new Oid(OID, detectionLocation));
+        this(PublicKeyEncryption.class, detectionLocation);
     }
 
     public RSA(int keyLength, @Nonnull DetectionLocation detectionLocation) {
-        super(NAME, PublicKeyEncryption.class, detectionLocation);
+        this(detectionLocation);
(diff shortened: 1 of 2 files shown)


san-zrl:
Test failed:
```
[ERROR] Failures:
[ERROR]   Issue127Test.test:37
Expecting actual:
  RSA
to be an instance of:
  com.ibm.mapper.model.algorithms.RSA
but was instance of:
  com.ibm.mapper.model.Algorithm
[ERROR]   JcaCipherMapperTest.pbe:82
expected: "pbeWithMD5AndDES"
 but was: "pbeWithMD5AndDES56"
[ERROR]   CipherSuiteMapperTest.test1:64
Expecting UnmodifiableRandomAccessList:
  [DH, AES, DSA]
to contain:
  [AES, DSA, SHA256, DH]
but could not find the following element(s):
  [SHA256]
```

PR #151, opened 2024-09-23
Fix issue report for python to SonarQube UI

n1ckl0sk0rtge:


hugoqnc (approved):
It fixed the issue: I tested that findings are correctly reported in the SonarQube UI after a scan

PR #152, opened 2024-09-23
SonarQube 10.6 support

hugoqnc:


docker-compose.yaml
```diff
@@ -1,6 +1,6 @@
 services:
   sonarqube:
-    image: sonarqube:10.4-community
+    image: sonarqube:10.6-community
```
n1ckl0sk0rtge, line 3: Didn't we want to use `latest` here?

n1ckl0sk0rtge (changes requested):

n1ckl0sk0rtge (approved):

PR #344, opened 2025-10-28
Fix: enrich asset collections

san-zrl:
Added enrichment for `AbstractAssetCollection`.


sonar-cryptography-plugin/src/main/java/com/ibm/plugin/Configuration.java
```diff
@@ -31,7 +31,7 @@ private Configuration() {}
     public static @Nonnull List<PropertyDefinition> getPropertyDefinitions() {
         return List.of(
                 PropertyDefinition.builder(Constants.CBOM_OUTPUT_NAME)
-                        .onQualifiers(Qualifiers.PROJECT)
```
n1ckl0sk0rtge: @san-zrl what is the reason for that change?

PR #362, opened 2026-01-25
Update algorithm names to match CycloneDX schema

n1ckl0sk0rtge:
## Summary

- Updates algorithm naming conventions in mapper model to align with the [CycloneDX cryptography-defs.json](https://cyclonedx.org/schema/cryptography-defs.json) specification
- Adds CycloneDX schema JavaDoc references to algorithm classes
- Updates test assertions across all language modules (Java, Python, Go)


bhess (commented):
Excellent!
Did you run an automatic check to verify that the patterns match?
I noticed a few that don’t, see the inline comments.

san-zrl:
What's the state of this? Will there be further changes?

n1ckl0sk0rtge:
Pushed follow-up commits addressing @bhess's review. I audited every algorithm class against the CycloneDX pattern format and added a dedicated unit test (`mapper/.../algorithms/AlgorithmNameCompositionTest`) that pins the composed name for each rule.

**Review concerns addressed:**

1. **KDF underscores** — the `SP800_108_*` family now uses underscores to match the registry:
   `SP800_108_CounterKDF`, `SP800_108_FeedbackKDF`, `SP800_108_DoublePipelineKDF`.

2. **PBES1** — `pbeWithHmacSHA-256AndAES` → `PBES1-AES-HMAC-SHA-256` (and `pbeWithMD5AndDES` → `PBES1-DES-56-MD5`).

3. **`XYZwithRSA` / `XYZwithDSA` patterns** — all *translated* (CBOM output) representations now follow the schema, e.g.:
   - `SHA256withRSA` → `RSA-PKCS1-1.5-SHA-256` (a plain RSA signature is PKCS#1 v1.5 by elimination, since PSS is `RSAssaPSS`/`RSA-PSS`)
   - `SHA384withDSA` → `DSA-2048-SHA-384`, `SHA-256withDSA` → `DSA-SHA-256`
   - `SHA-3-512withECDSA` → `ECDSA-secp384r1-SHA3-512`

   Every `// Noncompliant {{(Signature) …}}` annotation and translated-node assertion is converted; the raw `XYZwithRSA` strings only remain where they represent the **detected source** (`Signature.getInstance("…")` in analyzed files, mapper `parse(…)` inputs, and detection-store evidence values), which intentionally mirror what's written in the code rather than the CBOM name.

**Other pattern fixes in the same sweep:** `SHA-3` → `SHA3` (registry uses no hyphen, e.g. `SHA3-256`), `MQV` → `FFMQV`, `Aria`→`ARIA`, `Camellia`→`CAMELLIA`, `CAST-128/256`→`CAST5/6`, `SCRYPT`→`scrypt`, `Ascon-128`→`Ascon-AEAD128` (+`-Hash256`/`-XOF128`), `XMSS^MT`→`XMSSMT`, `CMAC-AES` ordering, and hyphenated composition (`AES-256-CBC`, `DES-56-CBC`, `CAST5-128`, `BLAKE2b-256`, `ChaCha20-Poly1305`).

Composition follows the registry rule of including only the parameters a pattern actually lists (e.g. `DES[-{keyLength}][-{mode}]` drops padding; `RC4[-{keyLength}]` is key-length only).

Full test suite is green across all modules.

> Did you run an automatic check to verify that the patterns match?

Not a formal validator against the schema regexes — the new `AlgorithmNameCompositionTest` asserts each expected composed string by hand. Happy to add a schema-regex conformance check if you'd like that as a guardrail.

n1ckl0sk0rtge:
Added the automatic schema-conformance check: `mapper/.../algorithms/AlgorithmSchemaConformanceTest`.

It translates each **verbatim CycloneDX pattern-format string** (e.g. `AES[-(128|192|256)][-(ECB|CBC|…)][-{padding}]`, `RSA-PKCS1-1.5[-{digestAlgorithm}][-{keyLength}]`, `SP800_108_(CounterKDF|FeedbackKDF|DoublePipelineKDF|KMAC)[-{prfFunction}][-{dkmLength}]`) into a regex — `[x]`→optional, `(a|b)`→alternatives, `{placeholder}`→a value regex, everything else literal — and asserts that each algorithm's composed `asString()` is a valid instance of its pattern. The translator itself has a small self-test (verifies optional/alternative groups, that the literal `.` in `1.5` isn't treated as a wildcard, and that a wrong prefix is rejected).

This complements `AlgorithmNameCompositionTest` (which pins exact strings): the composition test catches *value* drift, the conformance test catches *shape* drift against the registry patterns — e.g. it would now fail loudly if a signature regressed to `SHA-256-RSA` instead of `RSA-PKCS1-1.5-SHA-256`.

Covers ~35 families across block/stream/AEAD ciphers, hashes/XOFs, MACs, signatures & key exchange (RSA/DSA/ECDSA/ML-DSA/ML-KEM/FFMQV), and KDFs.

One intentional gap worth calling out: generic RSA **encryption** with only a detected key length emits the bare family token `RSA-2048` rather than asserting a padding scheme (`RSA-PKCS1-1.5-2048` / `RSA-OAEP-…`), since the scheme isn't detected there — so those aren't asserted against the RSA patterns. Happy to switch to always emitting a scheme if you'd prefer strict pattern conformance over "only detected parameters".

PR #388, opened 2026-05-02
Fixes issue of empty CBOM creation when no SonarQube rule is activated 

### LINKED ISSUE #265: Empty CBOM will beg generated even if no SonarQube rule is activated

medha-14:
fixes #265

### DIFF
--- a/sonar-cryptography-plugin/src/main/java/com/ibm/plugin/OutputFileJob.java
+++ b/sonar-cryptography-plugin/src/main/java/com/ibm/plugin/OutputFileJob.java
@@ -43,6 +43,11 @@ public class OutputFileJob implements PostJob {
                         .get(Constants.CBOM_OUTPUT_NAME)
                         .orElse(Constants.CBOM_OUTPUT_NAME_DEFAULT);
         ScannerManager scannerManager = new ScannerManager(new CBOMOutputFileFactory());
+        if (!scannerManager.hasResults()) {
+            LOGGER.info("No cryptography assets were detected. CBOM will not be generated.");
+            scannerManager.reset();
+            return;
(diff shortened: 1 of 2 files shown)


medha-14:
I’ve addressed the issue with the empty `cbom.json` creation. Please review it when you get a chance and let me know if any further changes are required.

n1ckl0sk0rtge (approved):
Thanks for the contribution, @medha-14! 🙌

LGTM!

One friendly nudge: if it's feasible, a small regression test asserting "no CBOM file is written when aggregators are empty" would lock this in. I noticed there aren't any existing tests for `OutputFileJob` / `ScannerManager` (the static aggregators make it a bit awkward to test cleanly), so this is a *nice-to-have* rather than a blocker — happy to merge as-is.

Approving. Thanks again!

n1ckl0sk0rtge (dismissed):

medha-14:
@n1ckl0sk0rtge Thanks for the review. I have added the tests, please have a look and let me know if anything else is needed. I’d be happy to iterate further on this if needed.

san-zrl (changes requested):
Hi @medha-14,

Thanks again for submitting this PR. Before merging could you remove the change of the `user.dir` system property in the test?

The test already uses an absolute path for the temporary folder

```
String cbomBaseName = tempDir.resolve("cbom").toAbsolutePath().toString();
```

Absolute paths don't depend on user.dir: When you pass an absolute path to new File(), Java doesn't use the working directory at all - it uses the path exactly as provided.

The user.dir manipulation is redundant. It is only necessary if relative paths were used.

san-zrl (approved):

PR #389, opened 2026-05-03
feat(pbes1): add OID mappings for PBES1 combinations

### LINKED ISSUE #384: Missing OID mappings in PBES1 algorithm model
**Description**

The `PBES1` class in `mapper/src/main/java/com/ibm/mapper/model/PBES1.java` does not resolve OIDs for any of its algorithm combinations. This means CBOM output for PBES1-based encryption will always be missing the `oid` field, making it non-compliant with the CycloneDX CBOM spec. 

**Expected behavior**

When a `PBES1` node has a known digest + cipher combination, the corresponding OID should be set automatically.

**Affected combinations (from RFC 2898 Appendix A.3 and PKCS#12)**

| Digest | Cipher       | Key length | OID                          |
|--------|--------------|------------|------------------------------|
| MD2    | DES-CBC      | —          | 1.2.840.113549.1.5.1         |
| MD5    | DES-CBC      | —          | 1.2.840.113549.1.5.3         |
| MD2    | RC2-CBC      | —          | 1.2.840.113549.1.5.4         |
| MD5    | RC2-CBC      | —          | 1.2.840.113549.1.5.6         |
| SHA1   | DES-CBC      | —          | 1.2.840.113549.1.5.10        |
| SHA1   | RC2-CBC      | —          | 1.2.840.113549.1.5.11        |
| SHA1   | RC4          | 128-bit    | 1.2.840.113549.1.12.1.1      |
| SHA1   | RC4          | 40-bit     | 1.2.840.113549.1.12.1.2      |
| SHA1   | TripleDES    | 3-key      | 1.2.840.113549.1.12.1.3      |
| SHA1   | TripleDES    | 2-key      | 1.2.840.113549.1.12.1.4      |
| SHA1   | RC2-CBC      | 128-bit    | 1.2.840.113549.1.12.1.5      |
| SHA1   | RC2-CBC      | 40-bit     | 1.2.840.113549.1.12.1.6      |

**References**
- RFC 2898 Appendix A.3: https://datatracker.ietf.org/doc/html/rfc2898#appendix-A.3
- PKCS#12 OIDs: https://www.alvestrand.no/objectid/1.2.840.113549.1.12.1.html

This issue was added as a TODO in the source code.
I am happy to work on a fix for this if the maintainers are open to it.

Ayush-Patel-56:
### Description
Added OID resolution for PBES1 algorithm combinations during the enrichment phase. This ensures CBOM output includes correct OIDs for both PKCS#5 and PKCS#12 variants, which were previously missing.

### Changes
- Created `PBES1Enricher` to map digest/cipher combinations to their respective OIDs (RFC 2898 and PKCS#12).
- Registered the new enricher in the `Enricher` pipeline.
- Removed legacy TODO from `PBES1.java`.
- Added `PBES1EnricherTest` with coverage for PKCS#5 and PKCS#12 scenarios.

### Related Issues
Fixes #384

### Verification
- Ran `mvn test -pl enricher -Dtest=PBES1EnricherTest`.
- Confirmed correct OID resolution for:
    - MD5 + DES (1.2.840.113549.1.5.3)
    - SHA1 + 3-Key TripleDES (1.2.840.113549.1.12.1.3)

### Screenshots
Attached is a screenshot showing the node tree before and after enrichment, confirming the OID is correctly resolved:
Screenshot 2026-05-03 204349

### DIFF
--- a/enricher/src/main/java/com/ibm/enricher/Enricher.java
+++ b/enricher/src/main/java/com/ibm/enricher/Enricher.java
@@ -26,6 +26,7 @@ import com.ibm.enricher.algorithm.DHEnricher;
 import com.ibm.enricher.algorithm.DSAEnricher;
 import com.ibm.enricher.algorithm.HMACEnricher;
 import com.ibm.enricher.algorithm.KEMEnricher;
+import com.ibm.enricher.algorithm.PBES1Enricher;
 import com.ibm.enricher.algorithm.PBKDF2Enricher;
 import com.ibm.enricher.algorithm.RSAEnricher;
 import com.ibm.enricher.algorithm.RSAoaepEnricher;
@@ -90,6 +91,7 @@ public class Enricher implements IEnricher {
                     new SHA3Enricher(),
                     new HMACEnricher(),
                     new PBKDF2Enricher(),
+                    new PBES1Enricher(),
                     new RSAssaPSSEnricher(),
                     new RSAoaepEnricher(),
                     new SignatureEnricher(),
--- a/enricher/src/main/java/com/ibm/enricher/algorithm/PBES1Enricher.java
+++ b/enricher/src/main/java/com/ibm/enricher/algorithm/PBES1Enricher.java
@@ -76,6 +76,7 @@ public class PBES1Enricher implements IEnricher {
             if (cipher instanceof RC4) {
                 if (keyLength == 128) return "1.2.840.113549.1.12.1.1";
                 if (keyLength == 40) return "1.2.840.113549.1.12.1.2";
+                // No PKCS#5 OID exists for SHA1+RC4 with other key lengths; null is intentional.
             }
             if (cipher instanceof TripleDES) {
                 if (keyLength == 192 || keyLength == 168) return "1.2.840.113549.1.12.1.3";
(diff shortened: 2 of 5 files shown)


n1ckl0sk0rtge:
@Ayush-Patel-56 Thanks for the PR!

The enricher approach is the right call and the implementation is clean. One small item to address before merge; everything else looks good.

### 1. RC4 silently drops on unknown key length (`PBES1Enricher.java:66`)
```java
if (cipher instanceof RC4) {
    if (keyLength == 128) return "1.2.840.113549.1.12.1.1";
    if (keyLength == 40) return "1.2.840.113549.1.12.1.2";
}
```
Unlike the RC2 branch, RC4 returns `null` with no fallback when the key length isn't 128 or 40. That's defensible (no PKCS#5 OID exists for SHA1+RC4), but the asymmetry vs. RC2 looks accidental. A one-line comment noting this is intentional would help future readers.

### Regression tests
Please add regression tests covering this case:
- SHA1 + RC4 with key length **not** 128/40 — asserts that no `Oid` child is attached (or whatever the resolved behavior ends up being).

A parameterized JUnit 5 test driven by the OID table in #384 would be the cleanest way to lock in all 12 rows at once and prevent future regressions, but at minimum the case above should have explicit coverage.

Ayush-Patel-56:
Fixed. Added a comment on the RC4 branch explaining the null return is intentional (no PKCS#5 OID exists for SHA1+RC4 with other key lengths). Expanded the test suite to a `@ParameterizedTest` covering all 14 OID rows plus a regression case asserting no `Oid` child is attached for SHA1+RC4 with key length outside 128/40.

san-zrl (approved):
@Ayush-Patel-56 - LGTM. Thank you for your contribution to CBOMkit. This is highly appreciated.

PR #390, opened 2026-05-04
fix: resolve intermediary variable detection in Java and Python engines

### LINKED ISSUE #8: Problem with handling detections of sub-rules
## Context

When writing a rule using the `DetectionRuleBuilder`, one can use the `shouldBeDetectedAs` function to resolve directly a specific parameter, and one can use the `addDependingDetectionRules` function to resolve a parameter with another rule (for example when this method is a method invocation, using a depending detection rule allows to resolve a specific parameter of this function call).
This issue is about the behaviour of the static analysis when resolving the following case (written here in Python, but the issue is also valid for Java):
```python
result = func1(func2("RSA")) # We want to resolve "RSA"
```

To do so, one should write a rule R1 to detect `func1`, with a depending detection rule R2 on its parameter to detect `func2`. Then R2 resolves the content of the parameter of `func2`, here "RSA", using `shouldBeDetectedAs`.
Currently (at main 0cd69ef79984b2c91bd0e06b07fb5488a2ea6473 and feature/python-support f5cdfe445fd4e5ea87ba6d0e886b84f02ad0c7cd), this type of resolution implies a complex mechanism that is quite limited. 

## Current mechanism

The detection process starts by visiting all method invocations (and class instantiations). For each invocation, a detection executive is started for each registered "entry" rule: R1 in our case.
We will therefore look for all functions `func1` in the current scope. Once we have a match, the function `analyseExpression` is called, and there are several cases:
- If the rule is only interested in the function without its parameter, it is handled in a special way that is not relevant for this issue.
- If the rule has a `DetectableParameter` (coming from `shouldBeDetectedAs`), meaning that we want to resolve a parameter of the detected function call, we name the subsequent handling logic *case of a detectable parameter*.
- If the rule has no `DetectableParameter` despite having parameters, it is probably a rule written for an intermediary function that does directly not carry information, but may have depending detections rules. We name the subsequent handling logic *case of an intermediary parameter*.

In our example, `func1` brings not information other than through the content of its parameter `func2("RSA")`. R1 is written as a rule with a depending detection rule R2 on its parameter, but without a `shouldBeDetectedAs` part.
We therefore enter the *case of an intermediary parameter* in the function `analyseExpression`.

### Case of an intermediary parameter

In this case, depending on the implementation, we have the choice to call `onDetectedDependingParameter` with two possible scopes: `EXPRESSION` or `ENCLOSED_METHOD`. Basically, it will resolve the depending detection rule(s) by visiting either the parameter tree (`func2("RSA")` in our case) or the enclosing method of this parameter (the function in which this expression is).

In our example, we want to resolve this particular `func2("RSA")` call, so using the EXPRESSION scope makes sense. `func2("RSA")` will be the only match of rule R2, which will correctly resolve the parameter content "RSA". This resolution is will go through the *case of a detectable parameter* explained below. This example worked as expected!

> **Problem 1**: This example worked because `func2("RSA")` was a function call. Here is a second example with an intermediary variable `var`:
> ```python
> var = func2("RSA")
> result = func1(var) # We want to resolve "RSA"
> ```
> In this case, because `var` is not a function call, visiting its tree will not return any results. Finding the function call associated to `var` is not that trivial, because there may be several intermediary assignments, results of function calls, dictionary values, etc.

### Case of a detectable parameter

Let's look at a new example:
```python
var = ec.ECDSA("SHA_256")
other_var = ec.ECDSA("SHA_512") # Some other var we are not interested in
result = func1(var) # We want to resolve "ECDSA" and "SHA_256"
```
In this example, `func1` still does not bring information other than through the content of its parameter `var = ec.ECDSA("SHA_256")`.
R1' (to detect `func1`) is written as a rule with a depending detection rule R2' for its parameter, but has now also a `shouldBeDetectedAs` part aiming at resolving the value "ECDSA" (name of the function used as parameter of `func1`).
R2' (to detect `ec.ECDSA`) (with `shouldBeDetectedAs`), resolves the content of the parameter of `ec.ECDSA`, here "SHA_256", using `shouldBeDetectedAs`.
Upon detection of `func1` with rule R1', because the parameter of `func1` is a detectable parameter, we enter the *case of a detectable parameter*.

Let's say that resolution works as expected in this case ("ECDSA" has been detected). Then, `analyseExpression` will call `onReceivingNewDetection`, that will call the following rules (here R2') with `followNextRules`. For now, there is no choice of scope, so it is necessarily the scope of the enclosing method. This means that it will look for a call of `ECDSA` in the entire scope of the method: how to make sure that this call is linked to our parameter `var` of `func1`?
To do so, the current implementation uses a **TraceSymbol**. `onReceivingNewDetection` will obtain the symbol of the parameter of `func1`, here `SYMBOL, var`, and will pass it as an argument to `followNextRules`.

Then, all function calls in the scope of the enclosing method will be matched against R2', which will detect the two `ECDSA` function calls, one assigned to `var` and the other to `other_var`. Using the previous TraceSymbol, we filter the findings to only keep the finding linked to `var`.
In our case, this works and this example will correctly resolve "SHA_256".

> **Problem 2**: This example worked because `ec.ECDSA("SHA_256")` was directly assigned to `var`. Here is a second example intermediary variable `intermediary_var`:
> ```python
> var = ec.ECDSA("SHA_256")
> intermediary_var = var
> other_var = ec.ECDSA("SHA_512") # Some other var we are not interested in
> result = func1(intermediary_var) # We want to resolve "ECDSA" and "SHA_256"
> ```
> In this case, `SYMBOL, intermediary_var` is the TraceSymbol obtained at the end of R1' and passed to `followNextRules`. But the symbol to which `ec.ECDSA("SHA_256")` is assigned is `var`. Because those two symbols don't match, there will be no detection. Again, correctly linking `intermediary_var` to `ec.ECDSA("SHA_256")` is not that trivial, because there may be several intermediary assignments, results of function calls, dictionary values, etc.

> **Problem 3**: What if there is no intermediary variable at all:
> ```python
> other_var = ec.ECDSA("SHA_512") # Some other var we are not interested in
> other_func(ec.ECDSA("SHA_512")) # Some other function call we are not interested in
> result = func1(ec.ECDSA("SHA_256")) # We want to resolve "ECDSA" and "SHA_256"
> ```
> In this case, `NO_SYMBOL` is the TraceSymbol obtained at the end of R1' and passed to `followNextRules`, because the parameter of `func1` is not a variable. Then, all function calls in the scope of the enclosing method will be matched against R2', which will detect the three `ECDSA` function calls, one assigned to `other_var` and the two others assigned to nothing (TraceSymbol `NO_SYMBOL`). The filtering step can then remove the `ECDSA` linked to the TraceSymbol `SYMBOL, other_var`, but it cannot distinguish between the two `NO_SYMBOL`. It will therefore resolve both SHA_512 and SHA_256.
> To avoid this case, we need to ensure that only the parameter of `func1` gets resolved. This could be done by limiting the scope to this parameter expression, as we can do in the *case of an intermediary parameter* (but we then need to solve Problem 1).

## A better approach

Could a better and simpler approach be found and replace this complex mechanism (currently based on limited TraceSymbols and varying scopes of resolution)?
> TODO: think about it.

Ayush-Patel-56:
# Description
Resolves #8 by implementing recursive symbol tracing in the Java and Python detection engines. 

The engine previously used shallow name-based matching for variables, which caused cryptographic rules to fail when intermediary variables were introduced (e.g., `SeatInterface intermediary = s; Car c = new Car(intermediary);`). This PR introduces a tracing mechanism that resolves variables back to their original initialization point.

# Changes
- **Recursive Tracing**: Added `traceSymbol` to follow variable assignments through the AST.
- **Deep Equivalence**: Implemented `areSymbolsEquivalent` to compare symbols after tracing them to their source declaration.
- **Engine Updates**: Modified `isInvocationOnVariable` and `isInitForVariable` in both `JavaDetectionEngine` and `PythonDetectionEngine` to use deep symbol resolution.
- **Regression Testing**: Added `Issue8IntermediaryVariableTest` covering nested rule detection across assignment levels.

# Verification
- **Unit Tests**: All 157 existing tests passed.
- **Manual Verification**: Confirmed via `DetectionStoreLogger` that the engine correctly associates children nodes through intermediary variables.
- **Trace Proof**: Logs confirm successful symbol resolution: `Comparing symbols: s (traced to s) with intermediary (traced to s)`.

*See attached screenshot for the successful test execution and detection hierarchy.*
Screenshot 2026-05-04 103511

### DIFF
--- a/engine/src/main/java/com/ibm/engine/language/java/JavaDetectionEngine.java
+++ b/engine/src/main/java/com/ibm/engine/language/java/JavaDetectionEngine.java
@@ -945,17 +945,25 @@ public final class JavaDetectionEngine implements IDetectionEngine<Tree, Symbol>
     @Nonnull
     private IdentifierTree traceVariable(@Nonnull IdentifierTree identifierTree) {
         Tree declaration = identifierTree.symbol().declaration();
-        if (declaration == null) {
+        if (declaration == null || !(declaration instanceof VariableTree variableTree)) {
             return identifierTree;
         }
-
-        if (declaration instanceof VariableTree variableTree1) {
-            ExpressionTree initTree = variableTree1.initializer();
-            if (initTree instanceof IdentifierTree identifierTree1) {
-                return traceVariable(identifierTree1);
-            }
+        ExpressionTree initTree = variableTree.initializer();
+        if (!(initTree instanceof IdentifierTree nextId)) {
+            return identifierTree;
         }
-        return identifierTree;
+        // Delegate chain-following to traceSymbol (single recursive walker)
+        Symbol finalSymbol = traceSymbol(nextId.symbol());
+        // Walk the identifier chain from nextId until we reach the identifier for finalSymbol
+        IdentifierTree current = nextId;
+        while (!current.symbol().equals(finalSymbol)) {
+            Tree decl = current.symbol().declaration();
+            if (!(decl instanceof VariableTree vt)) break;
+            ExpressionTree init = vt.initializer();
+            if (!(init instanceof IdentifierTree id)) break;
+            current = id;
+        }
+        return current;
     }
 
     /**
--- a/engine/src/main/java/com/ibm/engine/language/java/JavaDetectionEngine.java
+++ b/engine/src/main/java/com/ibm/engine/language/java/JavaDetectionEngine.java
@@ -691,7 +691,7 @@ public final class JavaDetectionEngine implements IDetectionEngine<Tree, Symbol>
 
             Symbol symbol = symbolOptional.get();
             if (symbol.isVariableSymbol()) {
-                return symbol.name().equals(variable.name());
+                return areSymbolsEquivalent(symbol, variable);
             }
             return true;
         }
@@ -713,11 +713,34 @@ public final class JavaDetectionEngine implements IDetectionEngine<Tree, Symbol>
 
             Symbol symbol = symbolOptional.get();
             if (symbol.isVariableSymbol()) {
-                return symbol.name().equals(variable.name());
+                return areSymbolsEquivalent(symbol, variable);
             }
             return true;
         }
         return false;
+    }
+
+    private boolean areSymbolsEquivalent(@Nonnull Symbol s1, @Nonnull Symbol s2) {
+        if (s1.equals(s2)) {
+            return true;
+        }
+
+        Symbol t1 = traceSymbol(s1);
+        Symbol t2 = traceSymbol(s2);
+
+        return t1.equals(t2);
+    }
+
+    @Nonnull
+    private Symbol traceSymbol(@Nonnull Symbol symbol) {
+        Tree declaration = symbol.declaration();
+        if (declaration instanceof VariableTree variableTree) {
+            ExpressionTree initializer = variableTree.initializer();
+            if (initializer instanceof IdentifierTree identifierTree) {
+                return traceSymbol(identifierTree.symbol());
+            }
+        }
+        return symbol;
     }
 
     @Nonnull
--- a/engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
+++ b/engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
@@ -18,7 +18,7 @@
  * limitations under the License.
  */
 package com.ibm.engine.language.python;
-
+ 
 import com.ibm.engine.detection.*;
 import com.ibm.engine.hooks.MethodInvocationHookWithParameterResolvement;
 import com.ibm.engine.hooks.MethodInvocationHookWithReturnResolvement;
@@ -34,6 +34,7 @@ import javax.annotation.Nullable;
 import org.sonar.plugins.python.api.PythonCheck;
 import org.sonar.plugins.python.api.PythonVisitorContext;
 import org.sonar.plugins.python.api.symbols.Symbol;
+import org.sonar.plugins.python.api.symbols.Usage;
 import org.sonar.plugins.python.api.tree.*;
 
 public class PythonDetectionEngine implements IDetectionEngine<Tree, Symbol> {
@@ -324,8 +325,10 @@ public class PythonDetectionEngine implements IDetectionEngine<Tree, Symbol> {
 
             QualifiedExpression qualifiedExpression = (QualifiedExpression) callee;
             if (qualifiedExpression.qualifier() instanceof Name name) {
-                Optional<String> nameString = Optional.of(name).map(Name::symbol).map(Symbol::name);
-                return nameString.isPresent() && nameString.get().equals(variable.name());
+                Symbol symbol = name.symbol();
+                if (symbol != null) {
+                    return areSymbolsEquivalent(symbol, variable);
+                }
             }
 
             return false;
@@ -346,7 +349,40 @@ public class PythonDetectionEngine implements IDetectionEngine<Tree, Symbol> {
 
         TraceSymbol<Symbol> traceSymbol = symbolOptional.get();
         Symbol symbol = traceSymbol.getSymbol();
-        return symbol.name().equals(variable.name());
+        if (symbol == null) {
+            return false;
+        }
+        return areSymbolsEquivalent(symbol, variable);
+    }
+
+    private boolean areSymbolsEquivalent(@Nonnull Symbol s1, @Nonnull Symbol s2) {
+        if (s1.equals(s2)) {
+            return true;
+        }
+
+        Symbol t1 = traceSymbol(s1);
+        Symbol t2 = traceSymbol(s2);
+
+        return t1.equals(t2);
+    }
+
+    @Nonnull
+    private Symbol traceSymbol(@Nonnull Symbol symbol) {
+        for (Usage usage : symbol.usages()) {
+            if (usage.kind() == Usage.Kind.ASSIGNMENT_LHS) {
+                Tree parent = usage.tree().parent();
+                if (parent instanceof AssignmentStatement assignment) {
+                    Expression rhs = assignment.assignedValue();
+                    if (rhs instanceof Name name) {
+                        Symbol rhsSymbol = name.symbol();
+                        if (rhsSymbol != null && !rhsSymbol.equals(symbol)) {
+                            return traceSymbol(rhsSymbol);
+                        }
+                    }
+                }
+            }
+        }
+        return symbol;
     }
 
     private void analyseExpression(
--- a/engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
+++ b/engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
@@ -18,7 +18,7 @@
  * limitations under the License.
  */
 package com.ibm.engine.language.python;
- 
+
 import com.ibm.engine.detection.*;
 import com.ibm.engine.hooks.MethodInvocationHookWithParameterResolvement;
 import com.ibm.engine.hooks.MethodInvocationHookWithReturnResolvement;
@@ -368,6 +368,10 @@ public class PythonDetectionEngine implements IDetectionEngine<Tree, Symbol> {
 
     @Nonnull
     private Symbol traceSymbol(@Nonnull Symbol symbol) {
+        // NOTE: symbol.usages() iteration order is not guaranteed. For a variable reassigned more
+        // than once (x = y; x = z; use(x)), this returns whichever ASSIGNMENT_LHS appears first in
+        // the iteration, which is non-deterministic. Ideally the assignment lexically nearest to
+        // the use site should be picked.
         for (Usage usage : symbol.usages()) {
             if (usage.kind() == Usage.Kind.ASSIGNMENT_LHS) {
                 Tree parent = usage.tree().parent();
--- a/python/src/test/files/rules/issues/Issue8IntermediaryVariableTestFile.py
+++ b/python/src/test/files/rules/issues/Issue8IntermediaryVariableTestFile.py
@@ -1,0 +1,9 @@
+from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
+
+key = b"\x00" * 32
+iv = b"\x00" * 16
+
+# intermediary variable: alg -> algorithms.AES(key)
+alg = algorithms.AES(key)
+intermediary = alg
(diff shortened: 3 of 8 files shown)


n1ckl0sk0rtge (changes requested):
@Ayush-Patel-56 Thanks for tackling this ❤️ 
The diagnosis is correct and the approach (replace name-equality with a recursive symbol trace) is reasonable. A few items below; flagging the bigger ones as required and one as a follow-up.

## Required before merge

**Correctness**

- **Python `traceSymbol` — first-usage-wins is non-deterministic.** The loop returns on the first `ASSIGNMENT_LHS` encountered while iterating `symbol.usages()`. For a variable reassigned more than once (`x = y; x = z; use(x)`), which RHS is followed depends on iteration order rather than which assignment reaches the call site. The original name comparison was wrong but predictable; this is wrong *and* non-deterministic. At minimum, document the limitation in a comment; ideally pick the assignment lexically nearest the use site.
- **C# engine not updated.** `CSharpDetectionEngine` has the same shallow `name().equals()` pattern. If C# is out of scope intentionally, say so in the PR description — otherwise apply the same fix there.

**Code quality**

- **Java duplicates an existing helper.** `JavaDetectionEngine.java:923` already has `traceVariable(IdentifierTree)` doing the same chain-of-initializers walk on `IdentifierTree`. The new `traceSymbol(Symbol)` is a near-clone. Refactor so there's one implementation (e.g., have `traceVariable` delegate to `traceSymbol`), rather than shipping two parallel walkers.
- **Stray whitespace in `PythonDetectionEngine.java:21`.** Blank line was changed to a line with a single space — unrelated to the fix. `mvn spotless:apply` should drop it.
- **`@Rule(key = "Issue8")` on the test class** isn't the convention here (see `Issue16Test`, `Issue214Test`, etc.). Please remove.
- **Trailing-whitespace-only lines** in `Issue8IntermediaryVariableTest.java`. Spotless will clean these up.
- **Unused `HeatedSeats` rule.** `seatRules` declares a rule for `HeatedSeats` but the test file never instantiates it. Drop the dead rule or add a `HeatedSeats` case.

**Test coverage**

The single new test covers the simplest case (`intermediary = s` as a constructor argument). For a change that introduces recursive resolution into both engines, please add:

- **Chain length > 1** (`a = s; b = a; new Car(b)`).
- **Method-invocation receiver path** (`isInvocationOnVariable`) — currently only `isInitForVariable` via `new Car(...)` is exercised.
- **At least one Python regression test.** The Python engine path has no test coverage in this PR, and Python's `usages()`-based trace differs structurally from Java's `declaration()`-based one — they need independent verification.

## Optional in this PR — please file a follow-up

- **Recursion guard.** Both `traceSymbol` methods recurse without a visited set or depth bound. Java is hard to construct a cycle in (compiler usually catches it); Python's `a = b; b = a` would loop forever, since the existing `!rhsSymbol.equals(symbol)` check only blocks the direct self-cycle. Real-world Python rarely produces this, so it's fine to defer — but please open a follow-up issue so it isn't lost. A small `Set<Symbol> visited` or a depth cap will close it cheaply.

Ayush-Patel-56:
Addressed in the latest commit:

- Added chain-length > 1 test case (`b → a → s`) to the existing intermediary variable test
- Added `Issue8MethodReceiverTest` to cover the `isInvocationOnVariable` path (`s.describe()` where `s` traces through an intermediary)
- Added Python `Issue8IntermediaryVariableTest` covering `Cipher(intermediary, ...)` resolving through a variable chain
Cleaned up the existing test - removed `@Rule`, `HeatedSeats`, and refactored `traceVariable` to delegate to `traceSymbol`

Will open a separate issue for the recursion guard. C# is intentionally out of scope as it doesn't have a `traceSymbol` equivalent yet.

Ayush-Patel-56:
Issue created: https://github.com/cbomkit/sonar-cryptography/issues/400

Ayush-Patel-56:
Opened #401 to add a recursion guard to `traceSymbol` for the mutual assignment cycle case (`a = b; b = a`) that scenario is not covered by this PR but is a natural edge case of the same tracing logic.

engine/src/main/java/com/ibm/engine/language/java/JavaDetectionEngine.java
```diff
@@ -713,13 +713,36 @@ public boolean isInitForVariable(Tree newClass, @Nonnull TraceSymbol<Symbol> var
...
+        }
+
+        Symbol t1 = traceSymbol(s1);
+        Symbol t2 = traceSymbol(s2);
+
+        return t1.equals(t2);
+    }
+
+    @Nonnull
+    private Symbol traceSymbol(@Nonnull Symbol symbol) {
```
n1ckl0sk0rtge, line 735: **Reassignment is ignored, and that turns a miss into a wrong answer.**

`traceSymbol` follows only the initializer of the declaration. Any later assignment to the same variable is invisible. So the trace can end at an object the variable no longer holds.

I ran this file against the rules from `Issue8IntermediaryVariableTest`:

```java
LeatherSeats s = new LeatherSeats();
SeatInterface a = s;
a = new HeatedSeats();   // 'a' now holds HeatedSeats
Car c = new Car(a);
```

| | child detections on `new Car(a)` |
|---|---|
| `main` today | 0 — nothing found |
| this PR | 1 — `LeatherSeats`, which is wrong |

Today the engine stays quiet. With this change it reports the wrong algorithm. For a CBOM a wrong entry is worse than a missing one, because nobody can tell it is wrong.

This is the Java twin of the Python ordering problem I raised last time, so please treat both the same way. Minimum: a comment here saying reassignment is not tracked, like the one you added on the Python side. Better: if the variable has more than one assignment, stop and return the original symbol instead of guessing.

engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
```diff
@@ -346,7 +349,44 @@ public boolean isInitForVariable(Tree newClass, TraceSymbol<Symbol> variableSymb
...
+        }
+
+        Symbol t1 = traceSymbol(s1);
+        Symbol t2 = traceSymbol(s2);
+
+        return t1.equals(t2);
+    }
+
+    @Nonnull
+    private Symbol traceSymbol(@Nonnull Symbol symbol) {
```
n1ckl0sk0rtge, line 370: **This method is never executed by any test in the repo.**

I put a `System.err.println` at the top of `areSymbolsEquivalent` and ran the new Python test. It prints nothing — zero calls.

I then ran the whole Python module (46 tests). The probe fires 8 times in total, all from existing key-agreement tests, and never with a chain that needs tracing:

```
5x areSymbolsEquivalent private_key vs private_key
1x areSymbolsEquivalent server_private_key vs server_private_key
1x areSymbolsEquivalent parameters vs server_private_key
1x areSymbolsEquivalent dh vs server_private_key
```

So `traceSymbol` on the Python side ships untested. Together with the ordering limit you documented in the note above, this is the part I am least comfortable merging. See my comment on the test file for what a real test needs.

python/src/test/files/rules/issues/Issue8IntermediaryVariableTestFile.py
```diff
@@ -0,0 +1,9 @@
+from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
+
+key = b"\x00" * 32
+iv = b"\x00" * 16
+
+# intermediary variable: alg -> algorithms.AES(key)
+alg = algorithms.AES(key)
+intermediary = alg
+cipher = Cipher(intermediary, modes.CBC(iv))  # Noncompliant
```
n1ckl0sk0rtge, line 9: **This test is green without the fix.**

I reverted `PythonDetectionEngine.java` to `main`, rebuilt the engine module, and ran this test alone (`mvn -pl python -am test -Dtest=Issue8*`):

```
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

So it does not cover the change. The Python engine already handles this case through another path. I also tried wrapping the code in a function, in case module-level scope was the reason. Same result: still green, still no call into the new code.

Compare with the two Java tests. Both fail without the engine change. That is what a regression test should look like.

To make this test real it has to reach `isInitForVariable` / `isInvocationOnVariable` with a `TraceSymbol` that does not match by name. The example from #8 is a good start, because the engine must pick the right one of two candidates:

```python
var = ec.ECDSA("SHA_256")
other_var = ec.ECDSA("SHA_512")   # decoy, must not be picked
intermediary_var = var
result = func1(intermediary_var)
```

Quickest way to know you got there: put a print in `areSymbolsEquivalent`, run the test, and check that it fires.

n1ckl0sk0rtge (changes requested):
Re-reviewed the latest push. I built the branch, ran the tests, and then reverted the two engine files to check that the new tests really go red.

Good news first:

- **The Java fix works.** Both Java tests fail without the engine change (`Expected size: 1 but was: 0`) and pass with it. That is a proper red-to-green test. Thank you for the method-receiver case and the longer chain.
- **No regressions.** The branch is 47 commits behind `main` but merges cleanly. I test-merged and ran the full `engine`, `java` and `python` suites on the merged result: all green.

Two things still block merge. Both are inline:

1. **The Python test does not test anything.** It is green with the Python change removed, and the new code is never called during the test.
2. **The Java trace ignores reassignment.** That turns a missed detection into a wrong detection, which is worse for a CBOM.

One small note: please also rebase on `main` before merge, and mention in the PR description that C# is out of scope (see below for why).

### FYI only — please do not add this to this PR

While checking the C# question I looked at the other two engines. Two separate gaps. I have opened an issue for each, so nothing here needs to grow this PR:

- **C# → #511**: `CSharpDetectionEngine.java:420` and `:439` still compare plain strings. This fix cannot be ported as-is, because `CSharpSymbol` holds only a name. The ANTLR grammar gives no `declaration()` and no `usages()` to walk. C# would need assignment tracking inside `CSharpTreeConverter`. So your "out of scope" call is right — the reason is architectural, and it is worth writing down.
- **Go → #510**: the opposite problem. `GoDetectionEngine.isInvocationOnVariable` and `isInitForVariable` both `return false` unconditionally (`:555`, `:565`), and `run(traceSymbol, tree)` ignores its `traceSymbol` argument. Go matches any call in the enclosing block with no variable check, so a child detection can come from an unrelated variable. 9 Go rule files use `addDependingDetectionRules`, so this is live today.

I confirmed both with a small repro before filing. The Go one is worse than it looks from the code: with two ECDSA keys in one function, one key raises no finding at all and the other is reported with the wrong curve.

PR #394, opened 2026-05-05
test(python): verify detection inside custom functions (part of #9)

sachin9058:
Adds a regression test to verify that cryptographic operations inside user-defined
functions are correctly detected during standard AST traversal.

### Motivation
While investigating issue #9, this PR validates that detection inside same-file
wrapper functions is already supported by the current visitor-based analysis.

This helps clarify that the remaining gaps in #9 are primarily related to
cross-file resolution and linking, rather than intra-file traversal.

### Changes
- Rename test to `PycaMacDetectionInCustomFunctionTest` for clarity
- Add negative test case to ensure non-cryptographic functions do not trigger findings
- Improve assertions and documentation for better readability and maintainability

### Scope
- No changes to detection logic
- Improves test coverage and documents existing behavior

### Example Covered

```python
def custom_sign(key, data):
    algorithm = hashes.SHA256()
    hmac_obj = hmac.HMAC(key, algorithm)  # detected
    hmac_obj.update(data)
    return hmac_obj.finalize()

### DIFF
--- a/python/src/test/files/rules/detection/mac/PycaMacDetectionInCustomFunctionTestFile.py
+++ b/python/src/test/files/rules/detection/mac/PycaMacDetectionInCustomFunctionTestFile.py
@@ -1,0 +1,27 @@
+from cryptography.hazmat.primitives import hmac
+from cryptography.hazmat.primitives import hashes
+
+def custom_sign(key, data):
+    # Custom function with cryptographic operation
+    algorithm = hashes.SHA256()
+    hmac_obj = hmac.HMAC(key, algorithm)  # Noncompliant {{(Mac) HMAC-SHA256}}
+    hmac_obj.update(data)
+    return hmac_obj.finalize()
+
+def non_crypto_function(text):
+    # Non-cryptographic function - should not trigger detection
+    result = "not crypto: " + text
+    return result.upper()
+
+# Example usage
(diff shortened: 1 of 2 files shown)


sachin9058:
@n1ckl0sk0rtge 

I investigated detection inside custom functions and found that intra-file traversal
is already handled by the current visitor pattern.

This PR adds a regression test to validate that behavior.

It seems the remaining challenges in #9 are primarily related to cross-file
resolution and linking of function calls.

sachin9058:
@n1ckl0sk0rtge 
I initially looked into implementing function-level resolution for this issue,
but during investigation I found that intra-file detection is already handled
by the current visitor-based traversal.

This PR adds a regression test to validate that behavior.

Given this, it seems the remaining challenges in #9 are primarily related to
cross-file resolution and linking of function calls, which is a broader
architectural concern.

I preferred to scope this contribution to validating existing behavior first
rather than introducing larger changes without full context.

sachin9058:
Hi @n1ckl0sk0rtge if u have a bit can u take a look on this pr ??

sachin9058:
@san-zrl Hi Andreas can u take a look on this PR ??

san-zrl (approved):
LGTM. Thank you very much for you contribution to cbomkit.

PR #397, opened 2026-05-05
feat: initial C++ support using standalone ANTLR scanner

Ndugu2:
Goal
This PR introduces initial support for C/C++ cryptographic detection, resolving Issue #374.

 Implementation Details
- Standalone Scanner**: Implemented a C++ sensor using standard ANTLR4 grammars (C++14), allowing for AST-based scanning without dependency on closed-source plugins.
- Detection Engine**: Added C++ specific translation and detection logic in the `engine` module.
- OpenSSL Rules**: Included an initial set of detection rules for the OpenSSL EVP API.
- Automated Source Generation**: Configured the `antlr4-maven-plugin` to generate parser classes at build-time. This PR correctly omits the massive generated Java files, keeping the repository hygiene high.

Cleanup & Hygiene
- This PR branch was cleaned to exclude unrelated C# scaffolding and binary artifacts (`maven_fast/`).
- Only source files and grammars are included.

Verification
- Includes `OsslEvpRuleTest.java` as a placeholder for architectural verification.
- To verify fully, run `mvn clean install` to ensure grammars compile and the module integrates correctly.

Fixes #374


san-zrl:
Hi @Ndugu2,

PR #402, opened 2026-05-06
feat(cpp): add CppCheck and CppSymbol engine types

Divyateja2709:
hi @n1ckl0sk0rtge

This PR adds the first C/C++ language-specific engine types required for upcoming C/C++ support in the Sonar Cryptography Plugin.

### Changes Included

- Added `CppCheck.java` as the marker interface for C/C++ detection rules
- Added `CppSymbol.java` as a lightweight symbol representation for C/C++ source identifiers
- Added support for optional statically resolved symbol values

### Purpose

The detection engine uses language-specific generic types for each supported language. These classes prepare the C/C++ support layer by introducing:

- `CppCheck` for the rule/check generic type
- `CppSymbol` for the symbol generic type

Because the planned C/C++ support is parser-based and does not yet include a full semantic model, `CppSymbol` intentionally keeps symbol handling lightweight.

This prepares future work on:

- C/C++ AST tree types
- C/C++ language translation
- C/C++ detection engine integration
- OpenSSL API detection rules

### DIFF
--- a/cpp/pom.xml
+++ b/cpp/pom.xml
@@ -1,0 +1,149 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ CBOMkit sonar-cryptography plugin
+  ~ Copyright (C) 2024 PQCA
+  ~
+  ~ Licensed to the Apache Software Foundation (ASF) under one or more
+  ~ contributor license agreements.  See the NOTICE file distributed with
+  ~ this work for additional information regarding copyright ownership.
+  ~ The ASF licenses this file to You under the Apache License, Version 2.0
+  ~ (the "License"); you may not use this file except in compliance with
+  ~ the License.  You may obtain a copy of the License at
+  ~
+  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


san-zrl:
Hi @Divyateja2709,

n1ckl0sk0rtge:
Hi @Divyateja2709,

Thank you again for this contribution and for the time you put into it.

We have decided to go with the approach in #377 for C/C++ support. It uses Sonar's native C/C++ analyzer instead of ANTLR, which fits the rest of the plugin better and gives us cross-file symbol tracing. So we are closing this PR.

A few notes on the current state of the branch, in case they help for future work:

- The build does not pass. `cpp/pom.xml` has parent version `1.5.1-SNAPSHOT`, but the root pom is `2.0.0-SNAPSHOT`. The module also declares the artifact id `sonar-cryptography-cpp`, while `sonar-cryptography-plugin/pom.xml` depends on `com.ibm:cpp`.
- The `cpp` module only has a pom. There is no language support class, scanner, or rule yet, but the plugin already lists `c,cpp,h` in `requiredForLanguages`.
- The grammar files sit under the `csharp` package path and `engine/pom.xml` does not run the ANTLR plugin, so nothing is generated from them. Since #496 we also load the C# parser/lexer as a published dependency instead of keeping grammars in this repo.

This is not a rejection of your work in general. We would be happy to see you contribute again, for example by helping review or extend #377, or by picking up one of the open issues. Thanks!

PR #403, opened 2026-05-06
feat(cpp): add base CppTree interface

Divyateja2709:
hi @n1ckl0sk0rtge

This PR adds the base `CppTree` interface for upcoming C/C++ language support.

### Changes Included

- Added `CppTree.java` under `engine/src/main/java/com/ibm/engine/language/cpp/tree/`
- Introduced the base tree type used as the engine's `T` generic parameter for C/C++
- Added basic node metadata methods for kind, line, and column

### Purpose

`CppTree` provides the shared base interface for future C/C++ AST node types such as literals, identifiers, function calls, member access, and blocks.

### DIFF
--- a/cpp/pom.xml
+++ b/cpp/pom.xml
@@ -1,0 +1,149 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ CBOMkit sonar-cryptography plugin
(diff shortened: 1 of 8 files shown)


san-zrl:
Hi @Divyateja2709,

PR #404, opened 2026-05-07
feat(cpp): add foundational C/C++ AST expression tree nodes

Divyateja2709:
## Summary

This PR adds foundational AST expression tree node implementations for the new C/C++ language support in the Sonar Cryptography Plugin.

The added nodes model core C/C++ expressions that are required for OpenSSL cryptographic API detection and future detection-engine traversal.

## Added Tree Nodes

* `CppLiteralTree`

  * Represents string and numeric literals
  * Examples:

    * `"AES-256-CBC"`
    * `"SHA256"`
    * `2048`

* `CppIdentifierTree`

  * Represents identifiers such as variable names, function names, and constants
  * Examples:

    * `ctx`
    * `EVP_aes_256_cbc`
    * `RSA_PKCS1_PADDING`

* `CppMemberAccessTree`

  * Represents member access expressions using `->` and `.`
  * Examples:

    * `ctx->cipher`
    * `params->key_len`
    * `config.algorithm`

## Purpose

These nodes provide the foundational expression-level AST representation required by the upcoming C/C++ detection engine implementation.

In particular:

* `CppIdentifierTree` enables function and variable name matching
* `CppLiteralTree` captures cryptographic constants and algorithm names
* `CppMemberAccessTree` supports object-type and struct-member resolution for OpenSSL APIs

This PR contains only AST data model additions and introduces no detection logic yet.

### DIFF
--- a/cpp/pom.xml
+++ b/cpp/pom.xml
@@ -1,0 +1,149 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ CBOMkit sonar-cryptography plugin
+  ~ Copyright (C) 2024 PQCA
+  ~
+  ~ Licensed to the Apache Software Foundation (ASF) under one or more
+  ~ contributor license agreements.  See the NOTICE file distributed with
+  ~ this work for additional information regarding copyright ownership.
+  ~ The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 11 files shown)


san-zrl:
Hi @Divyateja2709,

PR #408, opened 2026-05-10
feat(cpp): add C/C++ language translation, scan context, and base method visitor

Divyateja2709:
Description:
This PR adds the first three engine interface implementations for C/C++ language support. These are pure translation and context classes with no dependencies on each other, making them easy to review independently.Files added in engine/src/main/java/com/ibm/engine/language/cpp/:

- CppScanContext.java — fills the P generic slot. Wraps SonarQube's SensorContext and InputFile to report issues directly via the sensor API, same approach as C#.
- CppLanguageTranslation.java — implements ILanguageTranslation<CppTree>. Extracts method names, object type strings, and argument lists from CppMethodInvocationTree nodes. Since ANTLR4 has no type inference, all parameter types match any expected type.
- CppBaseMethodVisitor.java — implements IBaseMethodVisitor<CppTree>. When the sensor hands it a CppBlockTree (a function body), it runs the detection engine on it.

### DIFF
--- a/cpp/pom.xml
+++ b/cpp/pom.xml
@@ -1,0 +1,149 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ CBOMkit sonar-cryptography plugin
+  ~ Copyright (C) 2024 PQCA
+  ~
+  ~ Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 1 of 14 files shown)


san-zrl:
Hi @Divyateja2709,

Issue #31, opened 2024-07-02
BouncyCastle edge cases


hugoqnc:
- [ ] Investigate if capturing elliptic curves is possible
    - [ ] Handle cases where crypto assets have multiple nodes of the same type at the same level (like two digests)
    - [ ] Handle all static fields
    - [ ] Better handle enrichment: move all the enriched content currently in translation to enrichment

sachin9058:
@san-zrl Hi! I continued investigating BouncyCastle edge cases and opened a follow-up PR adding explicit regression coverage for additional digest implementations already registered through `BcDigests`.

Covered digests:

* WhirlpoolDigest
* TigerDigest
* RIPEMD160Digest

The focus is on improving explicit edge-case coverage and validating translation metadata without changing existing behavior.

PR #410, opened 2026-05-10
test(java): add RSADigestSigner constructor edge-case coverage

sachin9058:
This PR extends regression coverage for BouncyCastle signer edge cases related to `RSADigestSigner`.

### What changed

* Added a fixture covering the digest-only `RSADigestSigner` constructor path
* Extended the existing test expectations to validate the additional finding
* Verified the generated CBOM output remains consistent with the current RSA signature and digest relationship mapping

### Investigation Notes

While working on this issue, I also reviewed the existing EC/DSA signer paths:

* `ECDSASigner` support is already implemented through the existing `BcDSA` rule and mapper
* Existing test coverage for that path is already present
* `DSADigestSigner` currently resolves through the generic DSA translation flow, so I intentionally avoided adding misleading EC-specific assertions

This keeps the PR scoped specifically to improving regression coverage without changing current translation semantics.

### Validation

```bash id="smlqpd"
mvn -pl java -Dtest=BcRSADigestSignerTest test
mvn test -pl java
```

### Related

Closes part of #31


san-zrl (approved):
Hi @sachin9058 - LGTM. Thank you for your contribution to CBOMkit. This is highly appreciated.

PR #414, opened 2026-05-16
feat(engine): add C/C++ language support types and interfaces

medha-14:
Adds the foundational engine types for C/C++ language support, following
the pattern established by the C# module.


san-zrl:
Hi @medha-14,

You may have noticed that we are already working on adding C/C++ support to sonar-cryptography in PR #377. This approach relies on Sonar’s native C/C++ support package, similar to how the existing Java module is implemented.

PR #417, opened 2026-05-20
fix(engine, java): reduce heap retention by releasing AST and detection state eagerly

### LINKED ISSUE #371: High heap memory requirements

vdstech:
Closes #371

## Problem
Large Java codebases (28 000+ files) caused the scanner to consume up to
90 GB of heap. The root cause was `DetectionExecutive` holding strong
references to fully-resolved Java ASTs long after analysis completed,
combined with `DetectionStore` retaining its value/child collections for
the entire scan duration.

## Solution
Eagerly release ASTs and detection state after each file:

- **`DetectionExecutive`**: null out `tree` in a `finally` block in
  `start()` so the AST is GC-eligible immediately after rule analysis.
- **`DetectionExecutive`**: expose a deferred-hook lifecycle — when no
  deferred hooks are registered, `releaseResources()` is called eagerly
  after `emitFinding()`; when deferred hooks exist, `JavaBaseDetectionRule`
  drives cleanup via `releaseDeferredResources()` in `leaveFile()`.
- **`IStatusReporting`**: add `onDeferredHookRegistration()` default method
  as a backward-compatible lifecycle callback.
- **`DetectionStore`**: add `release()` to recursively clear
  `detectionValues`, `children`, and `actionValue`.
- **`JavaBaseDetectionRule`**: call `releaseDeferredResources()` on all
  deferred executives in `leaveFile()`.
- **`JavaScanMemoryLogger`**: new utility to log JVM heap usage every N
  files, making memory regressions visible in scan output.

## Testing
24 new unit tests covering the deferred-hook state machine, recursive
store release, and memory logger behaviour.

## Observed impact
Validated against Elasticsearch (28 000 Java files) and Kafka (6 000 Java
files). Peak heap dropped from >90 GB to normal scanner levels


n1ckl0sk0rtge:
## Review

Thanks for tackling #371 — the core memory optimization is sound, well-tested (24 new tests match the claim), and SPI-safe (default method on `IStatusReporting`, `tree` is `private` so the `@Nonnull` → `@Nullable` change is internal-only). Approval pending the items below + a performance validation artifact (see end).

### What this PR gets right (worth calling out)

- `tree = null` in `start()`'s `finally` — releases even on thrown AST analysis.
- `released` flag makes `releaseResources()` idempotent.
- `emitFinding()` wraps observer fan-out in `try` and releases only in `finally` — observers see live state, release runs after.
- `deferredDetectionExecutives` list cleared in `JavaBaseDetectionRule.leaveFile()`'s `finally`, so no per-scan growth.
- Tests cover idempotency, recursion, and both eager/deferred branches honestly.

### Should fix (non-blocking)

1. **Static logger state leaks across concurrent scans in the same JVM** — `JavaScanMemoryLogger.java:32-34`. `JAVA_FILES_PROCESSED`, `SUCCESSFUL_FILE_STATE_RESETS`, `PEAK_USED_MB` are static `AtomicLong`s. `JavaAggregator.reset()` clears them per project, but if two scans share a JVM (SonarLint daemon, parallel multi-module builds) counts and peak accumulate. Either scope the logger to a `JavaAggregator` instance or explicitly document the metric is JVM-global.

2. **`SUCCESSFUL_FILE_STATE_RESETS` counter is redundant** — `JavaScanMemoryLogger.java:45-46`. Both atomics are incremented unconditionally on the same line, so they always hold equal values, and `OutputFileJob` only prints `javaFilesProcessed`. Either remove the second counter (and drop the field from `Snapshot`) or wire it to a real reset signal.

3. **`DetectionStore.release()` recursion is unbounded** — `DetectionStore.java:262-269`. Post-order recursion is correct; on pathologically deep trees it could stack-overflow. Production ASTs sit well below the default ~5–10k stack-frame limit, so this is defensive only. Optional: convert to an iterative DFS with `ArrayDeque`.

4. **Same retention pattern likely exists in Python/Go/C# modules** — out of scope here, but worth filing a follow-up issue so the parity gap is tracked.

### Suggestions

5. **Redundant `tree = null` in `releaseResources()`** — `DetectionExecutive.java:153`. `start()`'s `finally` already nulls `tree`. Harmless, but signals confusion about ownership. Remove it, or add a one-line comment explaining the defensive intent.

6. **`IStatusReporting.onDeferredHookRegistration()` javadoc is thin** — `IStatusReporting.java:34`. \"Optional lifecycle callback for deferred hook-based findings.\" doesn't say when it fires (the `onNewHookRegistration` path), what implementations are expected to do (flip a flag that suppresses eager release), or that the default no-op preserves pre-PR semantics. Worth expanding for downstream language-module authors.

### Performance validation request

The PR description states \"Peak heap dropped from >90 GB to normal scanner levels\" on Elasticsearch (28k files) and Kafka (6k files), but no measurement artifacts are attached. Could you share a **JFR recording** comparing scan runs **with and without** this PR on the Elasticsearch codebase?

Suggested capture:

```bash
# Baseline: main branch
mvn org.sonarsource.scanner.maven:sonar-maven-plugin:sonar \
  -Dsonar.projectKey=elasticsearch \
  -DargLine=\"-XX:StartFlightRecording=duration=0,filename=/tmp/baseline.jfr,settings=profile\"

# This PR's branch (same -Xmx, same profile, same -DargLine but filename=/tmp/pr417.jfr)
```

Useful to see in the report:

- Peak `usedMb` from `CBOM memory progress` log lines (already emitted by the new logger — easy diff).
- JFR \"Java Application → Heap Configuration\" + \"Memory → Garbage Collections\" pages for both runs.
- A screenshot of \"Memory → Object Allocation\" filtered to `com.ibm.engine.detection.DetectionStore` and `org.sonar.java.model.JavaTree*` — that's the retention this PR targets, and JFR's allocation profiler shows it directly.
- Wall-clock scan duration delta (release/recursion has small overhead; want to confirm it isn't material).

Same `-Xmx` and same scan profile (`Inventory` + `JavaNoMD5use`, matching the issue reporter's setup) on both runs, so the diff is apples-to-apples.

vdstech:
## Performance Validation — JFR Memory Analysis
### PR #417: Fix DetectionStore recursive release + memory lifecycle
### Environment: -Xmx8g, same host, same sonar-project.properties
### All numbers extracted from JFR recordings via jdk.* events
---
## Primary Result
| Project       | Files  | Baseline          | PR #417          |
|---------------|--------|-------------------|------------------|
| Apache Kafka  | 2,915  | ❌ OOM at 49%     | ✅ Completed      |
| Elasticsearch | 3,648  | ❌ OOM at 12%     | ✅ Completed      |
---
## GC Heap Summary — all boundary measurements
### Apache Kafka
| GC | When      | No-fix (crashed) | With fix (done) | Delta    |
|----|-----------|------------------|-----------------|----------|
| 1  | Before GC | 23.66 MB         | 23.62 MB        | −0.04 MB |
| 1  | After GC  |  7.26 MB         |  7.19 MB        | −0.07 MB |
| 2  | Before GC | 79.26 MB         | 79.19 MB        | −0.07 MB |
| 2  | After GC  | 10.02 MB         |  9.63 MB        | −0.39 MB |
|    | Post GC 2 | 💥 OOM crash     | ✅ scan done     |          |
GC watermarks at GC 1 and GC 2 are virtually identical — confirming
both runs processed the same code under the same conditions. The
baseline collapsed after GC 2 when deep-tree files were reached.
### Elasticsearch
| GC | When      | No-fix (crashed) | With fix (done) | Delta     |
|----|-----------|------------------|-----------------|-----------|
| 1  | Before GC | 23.68 MB         |  19.91 MB       | −3.77 MB  |
| 1  | After GC  |  7.42 MB         |   8.73 MB       | +1.31 MB  |
| 2  | Before GC | 79.42 MB         |  56.73 MB       | −22.69 MB |
| 2  | After GC  |  9.90 MB         |  11.30 MB       | +1.40 MB  |
| 3  | Before GC | 💥 OOM crash     | 103.30 MB       | ← new GC  |
| 3  | After GC  | —                |  14.53 MB       |           |
The fix run reached a 3rd GC cycle that the baseline never got to —
processing the deepest-tree files (ComparisonTests.java: 24.5 min)
that caused the baseline OOM. GC 3 reclaimed 85.9% of heap in one
pass, proving the fix releases DetectionStore memory effectively.
---
## Peak Heap & GC Pressure
| Metric                  | Kafka no-fix | Kafka fix | ES no-fix | ES fix    |
|-------------------------|--------------|-----------|-----------|-----------|
| Total GC events         | 2 †          | 2         | 2 †       | 3         |
| Peak heap (before GC)   | 79.26 MB †   | 79.19 MB  | 79.42 MB †| 103.30 MB |
| Peak GCHeapMemoryUsage  | 70.02 MB     | 89.63 MB  | 69.90 MB  | 99.30 MB  |
| Avg heap after GC       | 8.64 MB      | 8.41 MB   | 8.66 MB   | 11.52 MB  |
| Old Gen (final)         | 3.26 MB      | 3.19 MB   | 3.42 MB   | 4.73 MB   |
| Survivor (final)        | 6.76 MB      | 6.44 MB   | 6.48 MB   | 9.81 MB   |
| Peak heap as % of -Xmx  | N/A (crashed)| 0.97%     | N/A       | 1.26%     |
† Partial recording — baseline crashed before scan completed.
The fix run ES peak (103 MB) is higher than no-fix partial peak
(79 MB) because the fix processed 3,648 files including the heaviest
crypto-analysis files the baseline never reached before crashing.
---
## RSS — Process Memory (jdk.ResidentSetSize)
| Run           | Peak RSS  | All-time peak |
|---------------|-----------|---------------|
| Kafka no-fix  | 195.66 MB | 205.45 MB     |
| Kafka fix     | 197.43 MB | 203.92 MB     |
| ES no-fix     | 194.11 MB | 203.12 MB     |
| ES fix        | 200.41 MB | 209.02 MB     |
RSS is consistent across all four runs (~195–210 MB). The OOM was
not caused by process-level memory growth — it was caused by heap
fragmentation from recursive DetectionStore trees preventing GC
from reclaiming live references. RSS confirms the fix adds no
process-level memory overhead.
---
## Thread Allocation (jdk.ThreadAllocationStatistics)
| Thread   | Kafka no-fix | Kafka fix | ES no-fix | ES fix    |
|----------|--------------|-----------|-----------|-----------|
| main     | 160.45 MB    | 168.05 MB | 159.75 MB | 162.65 MB |
| Thread-2 | —            | —         | —         | 25.72 MB  |
The `main` thread is the sole significant allocator in all runs.
`Thread-2` (25.72 MB) in the ES fix run is the deferred-hook
processing thread — enabled by the new deferredHookRegistered
lifecycle introduced in this PR. Its allocation is bounded and
correct: it represents real crypto findings being emitted from
deep-analysis files that the baseline never reached.
---
## Top Allocated Types (jdk.ObjectAllocationSample)
Consistent across all runs: primitive arrays ([J, [B, [I) and
JVM infrastructure (ConcurrentHashMap, ASM). BouncyCastle types
(ECFieldElement, ASN1Integer, ASN1ObjectIdentifier) are more
prominent in the fix runs because the fix actually completes
analysis of crypto-heavy files the baseline crashed before reaching.
No DetectionStore-related types appear in the top allocations —
confirming the fix prevents them from accumulating in heap.
---
## Baseline Crash Evidence
### Kafka — OOM at 49%
  Stalled at 49% for 80 seconds (19:11:31 → 19:12:51), then:
  OOM in: AST analyzer thread + main thread
  Slowest files (deep detection tree evidence):
    StandaloneHerderTest.java     684,884 ms   65 KB
    ConnectPluginPathTest.java    284,264 ms   32 KB
### Elasticsearch — OOM at 12%
  Stalled at 12% for 20+ minutes, elapsed 1h 24m 08s, then:
  OOM in: AST analyzer + HttpClient-1-SelectorManager + main
  Slowest files (deep detection tree evidence):
    ComparisonTests.java          1,472,155 ms  57 KB  ← 24.5 min
    AbstractSearchCancellationTestCase.java  232,801 ms
    SmokeTestWatcherTestSuiteIT.java         182,406 ms
---
## Attached Artifacts 
[summary.zip](https://github.com/user-attachments/files/28584146/summary.zip)

  kafka-no-fix/release.jfr       3.1 MB   partial (crashed at 49%)
  kafka-fix/pr417.jfr            8.5 MB   complete (2,915 files)
  es-no-fix/release.jfr          4.0 MB   partial (crashed at 12%)
  es-fix/pr417.jfr              37.6 MB   complete (3,648 files)
  JFR summary reports (text) also attached.
  No sensitive data present — verified via jdk.SystemProperty scan.

vdstech:
The 90 GB heap figure quoted in the PR description comes from our internal CI environment where all repositories belonging to a single product component are scanned together as one logical unit (effectively a single sonar-scanner invocation across tens of thousands of files). This is not an atypical single-project scan — it is a deliberate monorepo-style analysis that exercises the plugin at a scale most open-source users won't reach.

Because the environment runs on a spot instance, the allocation of sufficient heap for a baseline-vs-fix comparison run requires a formal capacity request. That ticket has been raised; approval and provisioning are pending.

In the interim, I am attaching Prometheus screenshots showing the RSS/heap trend during recent scans as supporting evidence of the memory pressure before this fix. A full JFR recording on the large internal codebase will be provided once the capacity ticket is resolved.

The JFR recordings already shared in this PR (Kafka, Elasticsearch) were captured on a locally available environment and demonstrate the same fix behaviour at smaller scale:

Kafka (6 000 files): OOM crash on baseline, clean completion on fix.
Elasticsearch (28 000 files): OOM crash on baseline, clean completion on fix.
These results are consistent with the internal observations and confirm the root cause and fix are correct. The large-scale JFR will follow as a final validation artifact.

Prometheus heap metrics from the internal scan environment are attached below for reference.
Without fix:
image

With Fix:
image

n1ckl0sk0rtge:
@vdstech Thanks a lot for digging into the heap problem — the eager release of hook-less executives is a real win and worth keeping. But two pieces of the cleanup assume hooks resolve *within a single file*, and they don't. Hooks are a **cross-file** mechanism, so this PR silently drops detections.

**How hooks actually work**

When a detection can't be resolved at its definition site, a lightweight hook (method signature + parameter index) is registered in the `HookRepository`. Take the `ResolveCallFromOtherClassTest` pattern:

```java
// defined in one class/file
public static Key generate(String algo, int keySize) {
    KeyGenerator keyGenerator = KeyGenerator.getInstance(algo); // algo is a parameter — unresolved here
    ...
}

// invoked from another class/file
Key test = ResolveCallFromOtherClassTestFile.generate("RSA", keySize); // Noncompliant {{(Key) RSA}}
```

The `getInstance(algo)` detection can't resolve `algo` at the definition, so a hook is registered. When an **invocation** of `generate(...)` is later encountered — in *any* file — resolution runs against that invocation's argument (`"RSA"`, `"AES"`) in the current file. `CallStackAgent` handles both orderings (retroactive scan in `onNewHookSubscription`, live firing via `addCall → notify`), which is what makes it independent of file scan order. The hook has to **survive across files** to match an invocation that lands in a different file than the definition.

**Where the PR breaks this**

1. **`JavaBaseDetectionRule.leaveFile()` → `JavaAggregator.resetLanguageSupport()`** recreates the entire `Handler`, wiping `HookRepository`, `HookDetectionObservable`, and `CallStackAgent`. The hook registered while scanning `generate`'s file is gone before `OtherClass`'s file is scanned, so the invocation never fires it.

2. **`leaveFile()` → `releaseDeferredExecutives()` → `releaseDeferredResources()`** releases the deferred executive at the *definition* file's `leaveFile`. But the invocation arrives in a later file. By then `rootDetectionStore.release()` has cleared the store the hook points at, and `listeners.clear()` has detached this rule — so even if the hook fired, the `(Key) RSA` / `(SecretKey) AES` finding couldn't propagate to `update()`. `leaveFile` of the registering file is the wrong release point for a cross-file hook.

**Why CI doesn't catch it**

`ResolveCallFromOtherClassTestFile.java` puts both classes in a **single file** — one compilation unit, one `leaveFile` — so it exercises cross-*class* resolution but never the cross-*file* reset path. There are no genuinely multi-file `CheckVerifier` tests in `java/src/test`, so this regression is invisible to a green build.

**The underlying tension**

The leak and the feature are the same retention: hook-subscribed stores + their trees stay alive across the scan precisely *because* they must remain matchable across files. So per-file reset/release can't be applied to hook state without losing detections.

**Suggested path**

- Keep the eager release of **non-deferred** (hook-less) executives + AST nulling — safe, and should recover the bulk of the heap.
- Drop `resetLanguageSupport()` per file and the `releaseDeferredResources()` call in `leaveFile()`, then re-measure.
- If hook/call-stack retention is still a material leak, address it with a strategy that preserves cross-file matchability (e.g. evict a hook + its store only once it's provably unmatchable, or retain a lighter representation of invocation arguments rather than whole subtrees).
- Please add a **truly multi-file** test — `generate(...)` in one `.java` file, the call site in a separate `.java` file — that asserts the cross-file detection still fires. Without it this regression stays invisible.

Happy to pair on the multi-file test harness if that helps.

n1ckl0sk0rtge:
see #371

PR #422, opened 2026-05-22
fix(csharp): implement translators for SignatureContext, PRNGContext, and ProtocolContext

pranjal2004838:
This PR completes the C# translation layer for cryptographic contexts that were left unimplemented in the initial C# support draft (PR #376, April 6, 2026).

Previously, CSharpTranslator returned Optional.empty() for SignatureContext, PRNGContext, and ProtocolContext, causing detected crypto assets (RSA signing, ECDSA, random number generation, TLS protocols) to be silently discarded from CBOM output.

**Changes:**
- CSharpSignatureContextTranslator: maps RSA, ECDSA, DSA to signature nodes
- CSharpPRNGContextTranslator: maps .NET random generators to PRNG nodes  
- CSharpProtocolContextTranslator: maps TLS and generic protocols to protocol nodes
- Added unit tests for each translator
- Updated CSharpTranslator dispatcher to route these contexts instead of discarding them

Fixes silent data loss in C# crypto detection pipeline.

### DIFF
--- a/csharp/src/main/java/com/ibm/plugin/translation/translator/CSharpTranslator.java
+++ b/csharp/src/main/java/com/ibm/plugin/translation/translator/CSharpTranslator.java
@@ -42,6 +42,9 @@ import com.ibm.plugin.translation.translator.contexts.CSharpCipherContextTransla
 import com.ibm.plugin.translation.translator.contexts.CSharpDigestContextTranslator;
 import com.ibm.plugin.translation.translator.contexts.CSharpKeyContextTranslator;
 import com.ibm.plugin.translation.translator.contexts.CSharpMacContextTranslator;
+import com.ibm.plugin.translation.translator.contexts.CSharpPRNGContextTranslator;
+import com.ibm.plugin.translation.translator.contexts.CSharpProtocolContextTranslator;
+import com.ibm.plugin.translation.translator.contexts.CSharpSignatureContextTranslator;
 import java.util.List;
 import java.util.Optional;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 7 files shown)


fynnth:
Hi @pranjal2004838,
Thanks for the contribution! Two things caught my eye that suggest this might not have been compiled and run (test pipeline fails because of unused imports, so i assume you have also not run it locally with mvn spotless):

1. Compilation error in CSharpSignatureContextTranslatorTest
```
SignatureAction<T> requires two constructor arguments -> Action and a location T. The test calls it with one:


SignatureAction<?> value = new SignatureAction<>(SignatureAction.Action.SIGN);
```
I believe this won't compile.

2. Contradictory assertion in testSignAction


```
assertThat(node.getKind()).isEqualTo(Signature.class);
assertThat(node.getKind()).isEqualTo(Sign.class);
```
Sign passes Sign.class to its super constructor, so getKind() returns Sign.class.  Both can't be true at the same time if im not wrong.

Did you run mvn test -pl csharp against these tests? I'd expect them to fail at build time.

**Aside from that, I want to flag some important context for this whole area:** I've temporarily excluded the C# module from main due to licensing issues with the ANTLR parser files that were included in the initial draft. I'll be writing a clean ANTLR grammar for C# from scratch, and once that's in place I also need to add more detection rules covering System.Security.Cryptography properly before the translation layer makes sense to extend.

Given that, I'd ask everyone to please hold off on C# enhancements for now and reach out to me first before putting work into this area. With the number of PRs coming in at the moment it's getting hard to keep track, and I'd hate for anyone to invest time in something that will conflict with the rewrite. Happy to coordinate once the grammar and detection rule foundation is solid.

pranjal2004838:
Great no issues, ill rather open it as an issue, if in future, the licensise problem of C# is resolved, do let me know, ill add these too.

Regards

Issue #339, opened 2025-10-01
Detection location 2 lines off


san-zrl:
When scanning `pkg:maven/com.google.guava/guava@33.0.0-jre` we get 12 findings all of which come from [Hashing.java](https://github.com/google/guava/blob/2214c63/guava/src/com/google/common/hash/Hashing.java). All findings refer to locations at the closing '*/' of javadoc comments. One such example is [Hashing.java#L364](https://github.com/google/guava/blob/2214c63/guava/src/com/google/common/hash/Hashing.java#L364-L364). 

We checked that the detection comes from the return line in the function body, not from the javadoc comment. This is correct but detection location is 2 lines off.

```java
/**
   * Returns a hash function implementing the Message Authentication Code (MAC) algorithm, using the
   * SHA-512 (512 hash bits) hash function and the given secret key.
   *
   * @param key the secret key
   * @throws IllegalArgumentException if the given key is inappropriate for initializing this MAC
   * @since 20.0
   */
public static HashFunction hmacSha512(Key key) {
  return new MacHashFunction("HmacSHA512", key, hmacToString("hmacSha512", key));
}
```

san-zrl:
The issue still persists. Scanning pkg:maven/com.google.guava/guava@33.0.0-jre produces 12 findings all of which come from [Hashing.java](https://github.com/google/guava/blob/2214c63/guava/src/com/google/common/hash/Hashing.java). The location of these findings are not correct but reported further down in the file than the actual function that is detected. The difference between actual and reported location is increasing.

1. Three MD5 detections (secret-key, md5, hmac-md5) for 
```java
return hmacMd5(new SecretKeySpec(checkNotNull(key), "HmacMD5"));
```
At the time of running the scan this was line 306 in Hashing.java . The detection is reported at line 330, 24 lines off.

2. Three SHA1 detections (secret-key, sha1, hmac-sha1) for 
```java
 return hmacSha1(new SecretKeySpec(checkNotNull(key), "HmacSHA1"));
```
At the time of running the scan this was line 330 in Hashing.java . The detection is reported at line 356, 26 lines off.

3. Three SHA256 detections (secret-key, sha256, hmac-sha256) for 
```java
 return hmacSha256(new SecretKeySpec(checkNotNull(key), "HmacSHA256"));
```
At the time of running the scan this was line 354 in Hashing.java . The detection is reported at line 382, 28 lines off.

[...]

PR #432, opened 2026-05-23
fix(java): propagate nested cryptographic values across method parameters

### LINKED ISSUE #431: Nested cryptographic detections are not propagated when cryptographic values are passed as method parameters
## Description

While investigating cryptographic detection propagation, I found that nested detections can be lost when cryptographic objects are passed as method parameters and later used inside another cryptographic construction.

The issue appears when analysis encounters parameter variables inside the callee method but does not resolve the original argument values from the caller invocation site.

As a result, cryptographic implementations passed by the caller are dropped before translation and enrichment, producing incomplete detection hierarchies.

---

## Reproduction

Example:

```java
public static byte[] calculateMac(
        BlockCipher cipher,
        int macSizeInBits,
        byte[] key,
        byte[] input) throws Exception {

    BlockCipherMac mac =
            new BlockCipherMac(
                    cipher,
                    macSizeInBits);

    ...
}

calculateMac(
        new AESEngine(),
        128,
        key,
        input);
```

The constructor receives:

```java
new BlockCipherMac(cipher, macSizeInBits)
```

where:

- `cipher` originates from `new AESEngine()`
- `macSizeInBits` originates from `128`

---

## Expected Behavior

The nested cryptographic implementation should be preserved.

Expected hierarchy:

```text
(MacContext) BlockCipherMac
└── (CipherContext) AESEngine
```

Additional parameter values should also remain available:

```text
(MacSize) 128
```

This preserves the full cryptographic context.

---

## Observed Behavior

Only the outer construction remains:

```text
(MacContext) BlockCipherMac
```

The nested cipher information is lost:

```text
AESEngine
```

does not appear in the final hierarchy.

---

## Investigation

During analysis, parameter variables appear to remain constrained to the local method scope.

When processing:

```java
BlockCipherMac(cipher, macSizeInBits)
```

the engine analyzes:

```java
cipher
macSizeInBits
```

inside the current method body but does not resolve them back to:

```java
calculateMac(
        new AESEngine(),
        128,
        key,
        input);
```

Because of this:

1. Caller values are not recovered
2. Nested detections are dropped
3. Translation and enrichment become incomplete
4. Final findings lose cryptographic context

---

## Impact

This may affect:

- Nested cryptographic detections
- Wrapper constructions
- Helper methods
- Factory methods
- Delegated cryptographic builders
- Method chains that pass algorithms through parameters

Potential consequences:

- Missing nested findings
- Incomplete cryptographic hierarchies
- Reduced reporting accuracy
- Loss of propagation across method boundaries

---

## Validation

A regression scenario was created using:

- `ParameterValuesNotCapturedTest`
- `ParameterValuesNotCapturedTestFile`

Validation scenario:

```java
calculateMac(
        new AESEngine(),
        128,
        key,
        input);
```

Expected output:

```text
BlockCipherMac
└── AESEngine
```

Current output:

```text
BlockCipherMac
```

---

## Possible Direction

When a tracked value represents a method parameter, analysis could resolve the corresponding invocation argument from the caller site before continuing detection.

Possible flow:

```text
caller value
      ↓
parameter resolution
      ↓
nested propagation
      ↓
translation
      ↓
final hierarchy
```

This would preserve cryptographic context across method boundaries and improve nested detection reporting.

sachin9058:
## Description

This PR resolves an issue where nested cryptographic detections are lost when cryptographic values are passed through method parameters.

During analysis, parameter variables were only evaluated inside the local method scope and the original values supplied by the caller were not propagated into the final detection hierarchy.

As a result, nested cryptographic implementations disappeared when helper methods, wrappers, or delegated constructions were used.

## Reproduction

Example:

```java
public static byte[] calculateMac(
        BlockCipher cipher,
        int macSizeInBits,
        byte[] key,
        byte[] input) throws Exception {

    BlockCipherMac mac =
            new BlockCipherMac(
                    cipher,
                    macSizeInBits);

    ...
}

calculateMac(
        new AESEngine(),
        128,
        key,
        input);
```

### Expected behavior

```text
(MacContext) BlockCipherMac
└── (CipherContext) AESEngine
```

Additional parameter values should also remain available:

```text
(MacSize) 128
```

### Previous behavior

```text
(MacContext) BlockCipherMac
```

The nested `AESEngine` detection was missing.

## Changes

- Added parameter value resolution for nested propagation
- Preserved caller values during analysis
- Updated parameter handling logic
- Added regression coverage:
  - `ParameterValuesNotCapturedTest`
  - `ParameterValuesNotCapturedTestFile`
- Updated nested hierarchy assertions

## Validation

Executed locally:

```bash
mvn -pl java test -Dtest=ParameterValuesNotCapturedTest
```

Result:

```text
Tests run: 1
Failures: 0
Errors: 0
Skipped: 0

BUILD SUCCESS
```

## Related issue

Fixes  #431

### DIFF
--- a/engine/src/main/java/com/ibm/engine/detection/DetectionStoreWithHook.java
+++ b/engine/src/main/java/com/ibm/engine/detection/DetectionStoreWithHook.java
@@ -149,8 +149,20 @@ public final class DetectionStoreWithHook<R, T, S, P> extends DetectionStore<R, 
                     detectionEngine.resolveValuesInInnerScope(Object.class, argument, null);
         }
         if (resolvedValues.isEmpty()) {
-            detectionEngine.resolveValuesInOuterScope(
-                    argument, methodInvocationHookWithParameterResolvement.getParameter());
+            if (!(methodInvocationHookWithParameterResolvement.getParameter()
+                            instanceof DetectableParameter<T>)
+                    && !methodInvocationHookWithParameterResolvement
+                            .getParameter()
+                            .getDetectionRules()
+                            .isEmpty()) {
+                hookRootDetectionStore.onDetectedDependingParameter(
+                        methodInvocationHookWithParameterResolvement.getParameter(),
+                        argument,
+                        DetectionStore.Scope.EXPRESSION);
+            } else {
+                detectionEngine.resolveValuesInOuterScope(
+                        argument, methodInvocationHookWithParameterResolvement.getParameter());
+            }
             return;
         }
 
--- a/engine/src/main/java/com/ibm/engine/detection/DetectionStoreWithHook.java
+++ b/engine/src/main/java/com/ibm/engine/detection/DetectionStoreWithHook.java
@@ -70,8 +70,7 @@ public final class DetectionStoreWithHook<R, T, S, P> extends DetectionStore<R, 
             // Add a hook to the hook repository
             if (handler.addHookToHookRepository(hook)) {
                 /*
-                 * Subscribes to the hook detection observable for the given hook value and
-                 * attaches the new
+                 * Subscribes to the hook detection observable for the given hook value and attaches the new
                  * Detection Store to it, so that it can receive detection events.
                  */
                 handler.subscribeToHookDetectionObservable(hook, hookRootDetectionStore);
@@ -222,8 +221,7 @@ public final class DetectionStoreWithHook<R, T, S, P> extends DetectionStore<R, 
                                     .visitMethodDefinition(hook.methodDefinition());
                         });
 
-        // add additional expected rule visits based on the size of the next detection
-        // rules
+        // add additional expected rule visits based on the size of the next detection rules
         statusReporting.addAdditionalExpectedRuleVisits(detectionRule.nextDetectionRules().size());
 
         detectionRule.nextDetectionRules().stream()
@@ -248,8 +246,7 @@ public final class DetectionStoreWithHook<R, T, S, P> extends DetectionStore<R, 
                                     .visitMethodDefinition(hook.methodDefinition());
                         });
 
-        // emit a finding to the status report if the root detection store contains any
-        // findings
+        // emit a finding to the status report if the root detection store contains any findings
         if (!isSuccessive) {
             statusReporting.emitFinding(hookRootDetectionStore);
         }
--- a/engine/src/main/java/com/ibm/engine/detection/DetectionStoreWithHook.java
+++ b/engine/src/main/java/com/ibm/engine/detection/DetectionStoreWithHook.java
@@ -70,7 +70,8 @@ public final class DetectionStoreWithHook<R, T, S, P> extends DetectionStore<R, 
             // Add a hook to the hook repository
             if (handler.addHookToHookRepository(hook)) {
                 /*
-                 * Subscribes to the hook detection observable for the given hook value and attaches the new
+                 * Subscribes to the hook detection observable for the given hook value and
+                 * attaches the new
                  * Detection Store to it, so that it can receive detection events.
                  */
                 handler.subscribeToHookDetectionObservable(hook, hookRootDetectionStore);
@@ -221,7 +222,8 @@ public final class DetectionStoreWithHook<R, T, S, P> extends DetectionStore<R, 
                                     .visitMethodDefinition(hook.methodDefinition());
                         });
 
-        // add additional expected rule visits based on the size of the next detection rules
+        // add additional expected rule visits based on the size of the next detection
+        // rules
         statusReporting.addAdditionalExpectedRuleVisits(detectionRule.nextDetectionRules().size());
 
         detectionRule.nextDetectionRules().stream()
@@ -246,7 +248,8 @@ public final class DetectionStoreWithHook<R, T, S, P> extends DetectionStore<R, 
                                     .visitMethodDefinition(hook.methodDefinition());
                         });
 
-        // emit a finding to the status report if the root detection store contains any findings
+        // emit a finding to the status report if the root detection store contains any
+        // findings
         if (!isSuccessive) {
             statusReporting.emitFinding(hookRootDetectionStore);
         }
--- a/engine/src/main/java/com/ibm/engine/language/java/JavaDetectionEngine.java
+++ b/engine/src/main/java/com/ibm/engine/language/java/JavaDetectionEngine.java
@@ -860,6 +860,20 @@ public final class JavaDetectionEngine implements IDetectionEngine<Tree, Symbol>
                     // scope
                     detectionStore.onDetectedDependingParameter(
                             parameter, newClassTree, DetectionStore.Scope.EXPRESSION);
+                } else if (expression
+                                instanceof
+                                org.sonar.plugins.java.api.tree.IdentifierTree identifierTree
+                        && identifierTree.symbol().isVariableSymbol()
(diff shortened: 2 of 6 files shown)


sachin9058:
Hi @san-zrl and @n1ckl0sk0rtge,

I investigated an issue where nested cryptographic detections were lost when cryptographic values were passed through method parameters.

Example:

`calculateMac(new AESEngine(), 128)`

which later becomes:

`new BlockCipherMac(cipher, macSizeInBits)`

Previously the nested value was not propagated and only `BlockCipherMac` appeared in the hierarchy. With this change the nested cipher (`AESEngine`) is preserved.

This PR includes:

* Parameter propagation changes
* Updates in the Java detection flow
* Regression coverage using `ParameterValuesNotCapturedTest`
* Validation for preserving nested detections across method boundaries

I would appreciate feedback on the approach and whether this direction aligns with the expected handling of parameter resolution.

Thanks!

sachin9058:
Hi @n1ckl0sk0rtge @san-zrl 
While validating the propagation changes I found an additional failure in `BcNTRUParametersTest` where `MessageDigest` appears during intermediate stages (translation/enrichment) but disappears before the final hierarchy assertion.

I opened #433 separately since it appears independent from parameter propagation and I wanted to keep this PR scoped to nested value propagation.

n1ckl0sk0rtge (changes requested):
## PR Review

**Linked Issue:** #431 — "Nested cryptographic detections are not propagated when cryptographic values are passed as method parameters" (OPEN)

| Area | Status |
|---|---|
| CI / Build | ❌ **FAILING** — `BcNTRUParametersTest` regression |
| Requirements coverage | Partial — fix appears to work for #431 example, but test skips an extra finding without justification |
| Architecture | OK — changes stay within `engine/language/java` |
| Scope | ❌ **CREEP** — bundles an unrelated fix for #339 (UI location) |
| Security | No concerns |
| Tests | Re-enables existing `ParameterValuesNotCapturedTest`; no test for the `JavaScanContext` changes |
| Breaking changes | Internal only |

---

### MUST FIX (Blocking)

**1. [Build] CI build is failing — `BcNTRUParametersTest` regression**
- `java/src/test/java/com/ibm/plugin/rules/detection/bc/cipherparameters/BcNTRUParametersTest.java:93`
- CI log shows `Expected size: 4 but was: 3` — the AES `BlockCipher` node now has only `{BlockSize, Oid, Encrypt}`, missing one child (most likely `KeyLength` or a mode marker).
- Likely root cause: the change in `JavaLanguageTranslation.java:201` removing `matchContext.isHookContext()` from the type-match condition alters how hook-context parameter types are matched, breaking the detection chain for `aesEngine.init(true, parameters)`.
- **Fix:** Reproduce locally with `mvn -pl java test -Dtest=BcNTRUParametersTest`, identify which child node was dropped, and either revert the `JavaLanguageTranslation` change or add the hook-context branch back with a comment explaining why it is no longer needed.

**2. [Scope] Bundled unrelated fix for issue #339**
- `engine/src/main/java/com/ibm/engine/language/java/JavaScanContext.java` (+54 lines — the biggest change in this PR)
- The `extractPreciseTree()` / `reportIssue()` refactor and the javadoc explicitly say *"fixes the SonarQube UI issue location for cases like #339"* — this is a separate concern from #431's parameter propagation.
- **Fix:** Split into a second PR linked to #339. Keeping unrelated changes together makes the regression in MUST-FIX #1 hard to bisect — the offending change could be in either the parameter logic or the new `extractPreciseTree` path.

### SHOULD FIX (Non-blocking)

**3. [Logic] `JavaLanguageTranslation.java:201` change is unjustified in the PR description**
- The diff drops `matchContext.isHookContext() ||` from the condition with no explanation. This is a behavior change for *every* hook-context match, not just the parameter-propagation path being fixed. Given finding #1, this is the most suspicious line in the PR.
- **Fix:** Add a code comment and a PR-body note explaining *why* hook context should no longer trigger exact matching, with a test that fails without the change and passes with it.

**4. [Tests] No coverage for the new `JavaScanContext` precise-location logic**
- Three new branches (`NEW_CLASS`, `METHOD_INVOCATION`, default fallback) are added with no unit test.
- **Fix:** Add a small test (e.g. `JavaScanContextTest`) that asserts the reported issue location for each case — `new AESEngine()` reports at `AESEngine`, `obj.method()` reports at `method`, other kinds fall back to `firstToken()`.

**5. [Tests] Extra finding skipped in `ParameterValuesNotCapturedTest` without explanation**
- `java/src/test/java/com/ibm/plugin/rules/issues/ParameterValuesNotCapturedTest.java:71` — changed from `findingId == 0` to `findingId == 0 || findingId == 2`. The existing TODO already calls out that these "direct detections of engines… appear in the depending detection rules"; adding finding 2 to the skip list quietly may mask a duplicate-detection regression introduced by the new parameter hook.
- **Fix:** In the test javadoc/comment, document what finding 2 represents and why it is expected to duplicate finding 1 (or fix the duplication if unintended).

**6. [Style] Fully-qualified class names instead of imports in `JavaDetectionEngine.java:863–879`**
- Uses `org.sonar.plugins.java.api.tree.IdentifierTree`, `…VariableTree`, `…MethodTree`, and `org.sonar.java.model.ExpressionUtils` inline. The rest of the file uses imports.
- **Fix:** Add normal imports and use simple names — consistent with file style and the project's Spotless/Checkstyle config.

### SUGGESTIONS

**7. [Heuristic] Parameter-detection heuristic in `JavaDetectionEngine.java:864`**
- `variableTree.initializer() == null` is meant to identify method parameters, but it also matches declared-but-not-yet-initialized locals (`String x; …`). For the next iteration consider `variableTree.parent().is(Tree.Kind.METHOD)` or checking the symbol's owner is a method, which is more precise.

**8. [Maintenance] `extractPreciseTree` duplicates token-selection logic**
- The javadoc says this mirrors `JavaTranslator.getDetectionContextFrom()`. Two parallel implementations will drift. Consider extracting to a single utility in `common/` (or referencing the existing one).

---

### Notes
- The fix idea for #431 (recognize parameter identifiers and re-enter analysis at the caller site via `createAMethodHook`) is sound and the re-enabled regression test demonstrates value.
- Please split the #339-related `JavaScanContext` changes into their own PR — keeping them here makes the failing `BcNTRUParametersTest` much harder to triage.
- Once the CI failure is investigated and #339 is split out, this PR will be in much better shape to merge.

sachin9058:
@n1ckl0sk0rtge Thanks for the detailed review.

I agree that the JavaScanContext changes are unrelated to #431 and should be split out. I'll remove the #339-related changes from this PR and keep the focus on parameter propagation.

I'll also investigate the BcNTRUParametersTest regression and review the JavaLanguageTranslation change around hook-context matching to determine whether it is responsible for the missing child node.

Thanks for pointing out the additional test/documentation gaps as well.

sachin9058:
Hi @n1ckl0sk0rtge,

I've addressed the review items:

* Fixed the BcNTRU regression by restricting caller-site hook creation to formal method parameters only.
* Restored the original `exactMatch` behavior in `JavaLanguageTranslation`.
* Removed the unrelated `JavaScanContext` (#339) changes from this PR.

Validation:

* `ParameterValuesNotCapturedTest` passes
* `BcNTRUParametersTest` passes
* Full test suite passes (`mvn clean test`)
* CI is green

Thanks for the review happy to make any further adjustments if needed.

PR #438, opened 2026-05-24
feat(cpp-openssl): add EVP cipher and digest detection rules

Rahulatram321:
## Summary
This PR adds initial OpenSSL EVP detection coverage focused on core encryption and digest entry points, enabling extraction of cryptography-relevant signals from common EVP usage patterns.

## Why
EVP is the primary high-level OpenSSL API used in production C/C++ code. Capturing EVP usage provides immediate high-value visibility into algorithms and operations.

## What this PR includes
- Detection rules for common EVP cipher and digest call patterns.
- Capture logic for:
  - Algorithm identifiers
  - Operation intent (encrypt/decrypt/hash where derivable)
  - Relevant invocation context metadata
- Rule registration under the C/C++ OpenSSL rule set.
- Baseline test fixtures for representative EVP usage examples.

## Architecture impact
- Introduces the first OpenSSL-specific rule bundle in C/C++ module.
- Uses existing rule-builder abstractions and context/bundle semantics.
- Keeps rule output in detection-store format for downstream translation.

## Validation
- Rule-level tests for positive detection on EVP flows.
- Negative tests for non-crypto or unrelated API calls.
- Verification that findings route through standard inventory path.

## Non-goals
- KDF, RNG, PKI, TLS coverage is out of scope for this PR.
- Full semantic translation mapping is completed in later translation-focused PRs.

## Follow-up
Next PRs expand API family coverage (KDF/RNG/PKI/TLS) and improve context-to-model mapping depth.

### DIFF
--- a/docs/mentorship/pr-03/README.md
+++ b/docs/mentorship/pr-03/README.md
@@ -1,0 +1,13 @@
+﻿# PR-3: feat(cpp-openssl): add EVP cipher and digest detection rules
+
+## Scope
+This branch contains the dedicated workstream for PR-3 under the C/C++ + OpenSSL mentorship track.
+
+## Deliverables
+- Clear implementation/tasks breakdown for this PR
+- Test and validation strategy
+- Integration notes with sonar-cryptography/CBOMkit
+- Risks, gaps, and follow-up actions
+
+## Status
+Initial structured content committed with DCO sign-off.


san-zrl:
Hi @Rahulatram321  - Our github PRs are a forum to discuss and approve code contributions. There are meant for planning. If you want to have a discussion on work streams and plans file an issue and provide an elaborate description on what you want to achieve, what use case would be addressed, why this is important and who would benefit.

PR #452, opened 2026-05-24
Fix algorithm recast child map copy

### LINKED ISSUE #451: Shallow copy in Algorithm recast constructor shares children map
## Bug

The `Algorithm(IAlgorithm algorithm, Class<? extends IPrimitive> asKind)` constructor copies metadata from an existing algorithm when recasting it as another primitive type, but it currently reuses the original algorithm's mutable children map.

**Current behavior:**

```java
this.children = algorithm.getChildren();
```

Because the same map reference is reused, any mutation to the recast node's children also mutates the original node.

## Example

One affected path is AES-GCM enrichment. `AESEnricher` adds an OID to the source AES node, then returns a recast AES node for authenticated encryption.

```java
aes.put(oid);
return new AES(AuthenticatedEncryption.class, aes);
```

With the current constructor, both AES objects share the same children map after recasting.

## Expected Behavior

The recast algorithm should receive an independent copy of the children map.

```java
this.children = new HashMap<>(algorithm.getChildren());
```

## Impact

This can affect every algorithm subclass that calls the shared recast constructor through `super(algorithm, asKind)`, including AES and other primitive model classes.

Any code that enriches or mutates the recast node can accidentally mutate the original algorithm node as well.

## Suggested Fix

Update `mapper/src/main/java/com/ibm/mapper/model/Algorithm.java` so the constructor copies the children map instead of assigning the same reference.

```java
public Algorithm(
        @Nonnull IAlgorithm algorithm, @Nonnull final Class<? extends IPrimitive> asKind) {
    this.name = algorithm.getName();
    this.children = new HashMap<>(algorithm.getChildren());
    this.detectionLocation = algorithm.getDetectionContext();
    this.kind = asKind;
    this.origin = algorithm.getOrigin();
}
```

## Suggested Regression Test

Add a test that verifies the original AES node is not modified when the returned recast node is mutated.

Test flow:

1. Create an AES-GCM algorithm node.
2. Enrich it with `AESEnricher`.
3. Confirm the enriched node is recast as `AuthenticatedEncryption`.
4. Mutate the returned node's children.
5. Assert that the original AES node still contains its original children.

This would confirm that recast algorithm instances no longer share mutable child state with the source algorithm.

Abhi190702:
## Summary

Fixes #451

This fixes a shallow-copy bug in the `Algorithm(IAlgorithm, asKind)` recast constructor.

The constructor previously reused the source algorithm's mutable children map, which meant the original algorithm node and the recast algorithm node shared the same child state.

## Root Cause

The recast constructor assigned the original map reference directly:

```java
this.children = algorithm.getChildren();
```

Because of this, mutating children on the recast node could also mutate the original algorithm node.

## Changes

- Copy the children map in `Algorithm(IAlgorithm, asKind)` using `new HashMap<>(algorithm.getChildren())`
- Add an AES-GCM regression test covering the recast path through `AESEnricher`
- Verify that mutating the returned/recast AES node does not remove children from the original AES node

## Validation

Ran the targeted regression test:

```bash
mvn -pl enricher -am -Dtest=AESEnricherTest -Dsurefire.failIfNoSpecifiedTests=false -Dexec.skip=true test
```

Result:

```text
Tests run: 4, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

### DIFF
--- a/enricher/src/test/java/com/ibm/enricher/algorithm/AESEnricherTest.java
+++ b/enricher/src/test/java/com/ibm/enricher/algorithm/AESEnricherTest.java
@@ -80,6 +80,25 @@ class AESEnricherTest extends TestBase {
     }
 
     @Test
+    void recastAesDoesNotShareChildrenWithOriginal() {
+        DetectionLocation testDetectionLocation =
+                new DetectionLocation("testfile", 1, 1, List.of("test"), () -> "SSL");
+        final AES aes = new AES(128, new GCM(testDetectionLocation), testDetectionLocation);
+
+        final AESEnricher aesEnricher = new AESEnricher();
+        final INode enriched = aesEnricher.enrich(aes);
+
+        assertThat(enriched).isInstanceOf(AES.class);
+        final AES enrichedAES = (AES) enriched;
+        enrichedAES.removeChildOfType(Oid.class);
+
+        assertThat(enrichedAES.hasChildOfType(Oid.class)).isEmpty();
+        assertThat(aes.hasChildOfType(Oid.class)).isPresent();
+        assertThat(aes.hasChildOfType(Oid.class).get().asString())
+                .isEqualTo("2.16.840.1.101.3.4.1.6");
+    }
+
+    @Test
     void defaultKeyLengthForJca() {
         DetectionLocation testDetectionLocation =
                 new DetectionLocation("testfile", 1, 1, List.of("test"), () -> "Jca");
--- a/mapper/src/main/java/com/ibm/mapper/model/Algorithm.java
+++ b/mapper/src/main/java/com/ibm/mapper/model/Algorithm.java
@@ -36,7 +36,7 @@ public class Algorithm implements IAlgorithm {
     public Algorithm(
             @Nonnull IAlgorithm algorithm, @Nonnull final Class<? extends IPrimitive> asKind) {
         this.name = algorithm.getName();
-        this.children = algorithm.getChildren();
+        this.children = new HashMap<>(algorithm.getChildren());
         this.detectionLocation = algorithm.getDetectionContext();
         this.kind = asKind;
         this.origin = algorithm.getOrigin();
--- a/mapper/src/main/java/com/ibm/mapper/model/Key.java
+++ b/mapper/src/main/java/com/ibm/mapper/model/Key.java
@@ -45,7 +45,7 @@ public class Key implements IAsset {
             @Nonnull DetectionLocation detectionLocation,
             @Nonnull final Class<? extends Key> asKind) {
         this.name = key.name;
-        this.children = key.getChildren();
+        this.children = new HashMap<>(key.getChildren());
         this.detectionLocation = detectionLocation;
         this.kind = asKind;
     }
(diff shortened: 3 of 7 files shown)


n1ckl0sk0rtge (changes requested):
Thanks for the fix! The change in `Algorithm.java` is correct and the regression test reliably catches the bug (verified locally: test passes with the fix, fails when the fix is reverted).

Before merging, please extend the scope — the **identical shallow-copy bug** exists in two sibling recast constructors:

### `mapper/src/main/java/com/ibm/mapper/model/Protocol.java:36`
```java
public Protocol(@Nonnull Protocol protocol, @Nonnull final Class<? extends Protocol> asKind) {
    this.type = protocol.type;
    this.children = protocol.getChildren();   // same bug
    ...
}
```

### `mapper/src/main/java/com/ibm/mapper/model/Key.java:48`
```java
protected Key(
        @Nonnull Key key,
        @Nonnull DetectionLocation detectionLocation,
        @Nonnull final Class<? extends Key> asKind) {
    this.name = key.name;
    this.children = key.getChildren();   // same bug
    ...
}
```

Both follow the exact pattern fixed in `Algorithm` and have the same failure mode: any consumer that recasts a `Protocol` or `Key` and mutates its children will silently mutate the source node. Linked issue #451 already calls out that "every algorithm subclass that calls the shared recast constructor" is at risk — the same logic applies to `Protocol`/`Key`. `HashMap` is already imported in both files, so the fix is the same one-liner: `new HashMap<>(source.getChildren())`.

Please also add regression tests for these two paths — a direct unit test on the constructor would be cleaner than going through enrichers (and would let you cover `Algorithm` more tightly too, decoupled from `AESEnricher`).

### Other notes

- **DCO check failing** — please amend with `git commit --amend --signoff` and force-push to add the `Signed-off-by:` trailer.
- **Informational (not blocking):** the fix is a map-level shallow copy — child `INode` references are still shared. That's correct for the reported bug (#451 is about `put`/`remove` mutations) and matches the existing semantics; `deepCopy()` already exists for full independence. Worth keeping in mind if future consumers mutate child state in place.

Otherwise looks good — fix is minimal, focused, and well-tested. Just want to close the same defect class in one PR rather than leaving the same bug live in two sibling classes.

Abhi190702:
Hey @n1ckl0sk0rtge !!! 

 Thanks for the detailed review!

I updated the sibling recast constructors in Protocol and Key to copy the children map with new HashMap<>(...), matching the existing Algorithm fix.

I also added direct regression tests for the Algorithm, Protocol, and Key recast constructor paths. Each test verifies that the recast node receives a separate children map instance and that mutating the recast map does not affect the original node's children map.

I amended the commit with --signoff and force-pushed to address the DCO failure.

Validation:

mvn -pl mapper -Dexec.skip=true test → 46 tests, 0 failures
mvn -pl enricher -am -Dtest=AESEnricherTest -Dsurefire.failIfNoSpecifiedTests=false -Dexec.skip=true test → 4 tests, 0 failures

Abhi190702:
hey @n1ckl0sk0rtge , Any progress ?

PR #457, opened 2026-05-29
Fix resolution of parameterized dictionary subscriptions

### LINKED ISSUE #10: Resolution for an outer parameter that is used in a subscription expression
Example inspired by [this code](https://github.com/ydb-platform/ydb/blob/284b7efb67edcdade0b12c849b7fad40739ad62b/contrib/python/Twisted/py2/twisted/conch/ssh/keys.py#L799).
```python
_curveTable = {
    b'ecdsa-sha2-nistp256': ec.SECP256R1(),
    b'ecdsa-sha2-nistp384': ec.SECP384R1(),
    b'ecdsa-sha2-nistp521': ec.SECP521R1(),
}

def _fromECComponents(cls, x, y, curve, privateValue=None):
        publicNumbers = ec.EllipticCurvePublicNumbers(
            x=x, y=y, curve=_curveTable[curve])
        #....

Key._fromECComponents(..., ..., ..., b'ecdsa-sha2-nistp256', ...) # Noncompliant {{SECP256R1}} (desired behaviour)
```

In this example, we want to resolve a curve value by looking into a dictionary. While this resolution is already implemented, this is a particular case where the _subscription index_ `curve` is a parameter of the enclosing function.
Here, `curve` is correctly resolved (using outer scope resolution) to `b'ecdsa-sha2-nistp256'`, but this resolved value is not later used to look into the dictionary.
Therefore, the captured value is currently `b'ecdsa-sha2-nistp256'` instead of `SECP256R1`.

somiljain2006:
Fixes #10 

Fix the resolution of dictionary subscriptions using function parameters. Preserve argument mappings when resolving subscription expressions and re-resolve outer-scope hook expressions using call-site arguments. This allows expressions such as _curveTable[curve] to resolve through dictionary lookups and correctly identify values like SECP256R1. Added a regression test.

### DIFF
--- a/engine/src/main/java/com/ibm/engine/detection/DetectionStoreWithHook.java
+++ b/engine/src/main/java/com/ibm/engine/detection/DetectionStoreWithHook.java
@@ -139,14 +139,35 @@ public final class DetectionStoreWithHook<R, T, S, P> extends DetectionStore<R, 
         }
 
         List<ResolvedValue<Object, T>> resolvedValues;
+        T expressionToResolve = methodInvocationHookWithParameterResolvement.expressionToResolve();
         if (methodInvocationHookWithParameterResolvement.getParameter()
                 instanceof DetectableParameter<T> detectableParameter) {
-            resolvedValues =
-                    detectionEngine.resolveValuesInInnerScope(
-                            Object.class, argument, detectableParameter.getiValueFactory());
+            if (expressionToResolve != null) {
+                resolvedValues =
+                        detectionEngine.resolveValuesInInnerScope(
+                                Object.class,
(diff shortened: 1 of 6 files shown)


somiljain2006:
@n1ckl0sk0rtge Can you review this pr?

san-zrl (approved):
- Fixes resolution of parameterized dictionary subscription #10
- Correctly resolves dictionary values (not keys) when parameters are used as indices
- Works with any subscription index type (string, int, bytes) that can be resolved to a literal
- All tests pass in CI and locally (after proper clean build)
- Adds appropriate test coverage for the new functionality

Issue #32, opened 2024-07-02
Use various providers in JCA


hugoqnc:
JCA uses a provider-based architecture, where one can specify a cryptography provider (like BouncyCastle) which will provide the implementations of JCA APIs.
- [ ] Investigate if we can often identify the provider when tracking JCA crypto calls
- [ ] If so, restructure the JCA rules to make them depend on the provider
- [ ] Add BouncyCastle as a cryptography provider for JCA

SurbhiAgarwal1:
I would like to work on this issue.

Proposed approach:

I think this can be implemented in multiple stages to keep the provider-aware architecture extensible and avoid tightly coupling rules directly to one implementation.

1. Provider Identification Layer
   Introduce a lightweight provider-resolution component for JCA analysis that attempts to determine the active cryptography provider from:

* Explicit provider arguments in calls such as:

  * `Cipher.getInstance(..., provider)`
  * `MessageDigest.getInstance(..., provider)`
  * `Signature.getInstance(..., provider)`
* Provider registration flows:

  * `Security.addProvider(...)`
  * `Security.insertProviderAt(...)`
* Fully qualified provider classes (for example BouncyCastle provider initialization).

This layer could expose normalized provider metadata such as:

```java
enum JcaProvider {
    DEFAULT,
    BOUNCY_CASTLE,
    SUN_JCE,
    UNKNOWN
}
```

2. Provider-Aware Rule Architecture
   Instead of having completely generic JCA rules, introduce a provider abstraction so rules can optionally specialize behavior depending on the detected provider.

For example:

* Generic JCA rule interface
* Shared algorithm resolution utilities
* Provider-specific extensions

This would help avoid false positives/negatives where algorithms, modes, padding support, or defaults differ across providers.

3. BouncyCastle Support
   Add BouncyCastle as the first provider-specific implementation because:

* it is widely used in enterprise/security-focused Java applications
* it exposes algorithms not available in default providers
* many cryptographic misuse patterns are provider-specific

Potential additions:

* BC-specific algorithm mapping
* insecure/deprecated BC algorithm detection
* provider-specific compliance/security checks

4. Dataflow & Symbol Analysis
   To improve provider detection accuracy, provider information should propagate through symbolic execution/dataflow when possible.

Example:

```java
Provider bc = new BouncyCastleProvider();
Cipher.getInstance("AES/GCM/NoPadding", bc);
```

or:

```java
Security.addProvider(new BouncyCastleProvider());
```

[...]

PR #458, opened 2026-06-01
feat(java): propagate JCA provider metadata to CBOM output

sachin9058:
## Summary

This PR adds support for capturing and propagating JCA provider metadata through the analysis pipeline and into CBOM output.

Provider information supplied to JCA APIs such as:

```java
Cipher.getInstance("AES/CBC/PKCS5Padding", "BC");
MessageDigest.getInstance("SHA-256", "SUN");
```

was previously accepted during rule matching but discarded before translation and CBOM generation.

This change preserves provider information throughout the pipeline and emits it as a CBOM component property.

### Engine

Added a new provider detection model:

* `Provider`
* `ProviderFactory`

The factory resolves provider names from String-based provider arguments and creates provider detections that can be attached to existing JCA detections.

### Detection Rules

Updated JCA `getInstance()` rules to capture provider parameters and attach them to the corresponding algorithm detection:

* Cipher
* MessageDigest
* Mac
* Signature
* KeyAgreement
* KeyFactory
* SecretKeyFactory
* KeyGenerator
* KeyPairGenerator
* AlgorithmParameterGenerator

Provider detections are attached as child detections of the primary algorithm parameter.

### Translation

Added provider translation support across all relevant JCA context translators.

A new mapper-level `Provider` model is introduced to carry provider metadata through the translation layer.

### CBOM Output

Provider metadata is now emitted as a CycloneDX component property:

```json
{
  "name": "jca-provider",
  "value": "BC"
}
```

The output builder was updated to support provider metadata and to correctly propagate providers for key-related JCA detections.

## Example

Input:

```java
Cipher.getInstance("AES/CBC/PKCS5Padding", "BC");
```

Output:

```json
{
  "properties": [
    {
      "name": "jca-provider",
      "value": "BC"
    }
  ]
}
```

## Testing

Updated and extended tests to validate provider propagation:

* `BcSM2EngineTest`
* `JcaMessageDigestGetInstance2Test`

Verification performed:

* All Java tests pass
* Full project build succeeds

## Notes

This PR focuses on provider detection, propagation, and CBOM representation.

Provider-aware rule behavior and provider-specific analysis can be implemented as follow-up work on top of this foundation.

Related to #32

### DIFF
--- a/engine/src/main/java/com/ibm/engine/model/Provider.java
+++ b/engine/src/main/java/com/ibm/engine/model/Provider.java
@@ -1,0 +1,68 @@
+/*
+ * Sonar Cryptography Plugin
+ * Copyright (C) 2024 PQCA
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.ibm.engine.model;
+
+import javax.annotation.Nonnull;
+
+public class Provider<T> extends AbstractValue<T> {
+    @Nonnull private final String value;
+    @Nonnull private final T location;
+
+    public Provider(@Nonnull String value, @Nonnull T location) {
+        this.location = location;
+        this.value = value;
+    }
+
+    @Nonnull
+    public String get() {
+        return value;
+    }
+
+    @Override
+    @Nonnull
+    public T getLocation() {
+        return location;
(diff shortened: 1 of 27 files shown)


sachin9058:
Hi @n1ckl0sk0rtge @n1ckl0sk0rtge @hugoqnc,,

I've opened a PR for issue #32 that implements provider metadata propagation for JCA `getInstance()` calls.

The change captures provider information (e.g. `"BC"` and `"SUN"`), propagates it through the engine and translation layers, and emits it in the generated CBOM as a `jca-provider` property.

This PR focuses on provider detection and propagation only. Provider-aware rule behavior can be built on top of this foundation in follow-up work.

All Java tests pass and the full project build succeeds.

I'd appreciate a review when you have time. Thanks!

sachin9058:
Thanks for the review.

I've addressed the suggestions:

* Updated the implementation to preserve existing component properties when adding the `jca-provider` property.
* Extracted the property name into a constant to avoid introducing a magic string.
* Added output-level test coverage (`AlgorithmTest`) to verify that provider metadata is emitted correctly in the generated CBOM output and that the expected property value is preserved.

All tests pass after the changes.

san-zrl:
Hi @sachin9058 - I see the point of this PR since it relates to an old issue. The primary goal of CBOMkit is to produce meaningful CBOMs that can be consumed by interested parties. The is not standard attribute we can possibly yse to store the JCA provider information. Putting it in the properties key-value table defeats the purpose; this creates our own specific dialect. Consumers of this would have to _know_ where the provider information is and how it must be used. For this reason I'd like to postpone the decision if at all we want to include JCA provider data and if so where and how.

PR #460, opened 2026-06-05
fix: detect pyca stream ciphers with mode=None

Truty424:
The existing block cipher detection rule requires the mode parameter to match `cryptography.hazmat.primitives.ciphers.modes.*`.

Stream ciphers such as `ChaCha20` do not use a cipher mode and are initialised with `mode=None`, so they were not detected.

This PR adds a detection rule for stream ciphers initialised with `mode=None`.

A test file and unit test were added for `Cipher(algorithms.ChaCha20(...), mode=None)`. The test passes successfully.


san-zrl (approved):
LGTM

PR #465, opened 2026-06-08
Fix asset deduplication logic

### LINKED ISSUE #295: Duplicate findings ?
Hello, I have been conducting experiments with your plugin and would like to share some of my results. Thank you for the great work!

### Context:
sonar-cryptography plugin version: 1.4.5

### Issue found:
Although sonar-cryptography does identify correctly AES usage, these findings look like duplicates:
I suppose this is part of the policy facilitating design choices where AES is reported separately because it is  a subcomponent of AES-CTR. But given that the same OID is used, does this choice really make sense ?

- [Code snippet](https://github.com/hyperchain/javasdk/blob/93b6ca10e5835160a255290fe47a1640295a4118/src/main/java/cn/hyperchain/sdk/crypto/ecdsa/ECKey.java#L947)

Finding 1:
```
{
      "type" : "cryptographic-asset",
      "bom-ref" : "fb749483-ba6a-44d4-af57-e04b1087e820",
      "name" : "AES",
      "evidence" : {
        "occurrences" : [
          {
            "location" : "src/main/java/cn/hyperchain/sdk/crypto/ecdsa/ECKey.java",
            "line" : 947,
            "offset" : 31,
            "additionalContext" : "org.bouncycastle.crypto.engines.AESFastEngine#<init>()V"
          }
        ]
      },
      "cryptoProperties" : {
        "assetType" : "algorithm",
        "algorithmProperties" : {
          "primitive" : "block-cipher",
          "parameterSetIdentifier" : "128"
        },
        "oid" : "2.16.840.1.101.3.4.1"
      }
    }
```

Finding 2:
```
{
      "type" : "cryptographic-asset",
      "bom-ref" : "a2be8a25-61b3-4d64-b14c-3557e964eb23",
      "name" : "AES-CTR",
      "evidence" : {
        "occurrences" : [
          {
            "location" : "src/main/java/cn/hyperchain/sdk/crypto/ecdsa/ECKey.java",
            "line" : 947,
            "offset" : 31,
            "additionalContext" : "org.bouncycastle.crypto.engines.AESFastEngine#<init>()V"
          }
        ]
      },
      "cryptoProperties" : {
        "assetType" : "algorithm",
        "algorithmProperties" : {
          "primitive" : "block-cipher",
          "parameterSetIdentifier" : "128",
          "mode" : "ctr",
          "cryptoFunctions" : [
            "decrypt"
          ]
        },
        "oid" : "2.16.840.1.101.3.4.1"
      }
    }
```

somiljain2006:
Fixes #295

When processing certain Bouncy Castle block cipher constructions, multiple block-cipher-related nodes can be produced for the same source location. The existing deduplication logic relied on strict checks against the BlockCipher interface, which prevented some algorithm nodes from participating in reorganization even when they represented block-cipher assets.

This change updates the deduplication rule to operate on Algorithm nodes while preserving the existing BlockCipher node-kind filtering. It also adds null-safe handling for algorithm names and improves merging behavior when generic wrapper nodes are encountered.

Tests were added to cover:
- overlapping block cipher roots generated from the Bouncy Castle engine and mode mappings
- nested parent/child block cipher structures processed by the reorganizer

### DIFF
--- a/java/src/main/java/com/ibm/plugin/translation/reorganizer/JavaReorganizerRules.java
+++ b/java/src/main/java/com/ibm/plugin/translation/reorganizer/JavaReorganizerRules.java
@@ -45,6 +45,7 @@ public final class JavaReorganizerRules {
                 AeadBlockCipherReorganizer.MOVE_TAG_LENGTH_UNDER_MAC,
                 AsymmetricBlockCipherReorganizer.INVERT_DIGEST_AND_ITS_SIZE,
                 AsymmetricBlockCipherReorganizer.MERGE_PKE_PARENT_AND_CHILD,
+                BlockCipherReorganizer.DEDUPLICATE_OVERLAPPING_ROOTS,
                 BlockCipherReorganizer.MERGE_BLOCK_CIPHER_PARENT_AND_CHILD,
                 CipherParameterReorganizer.MOVE_KEY_LENGTH_UNDER_TAG_LENGTH_UP,
                 CipherParameterReorganizer.MOVE_NODES_UNDER_DECRYPT_UP,
--- a/mapper/src/main/java/com/ibm/mapper/reorganizer/rules/BlockCipherReorganizer.java
+++ b/mapper/src/main/java/com/ibm/mapper/reorganizer/rules/BlockCipherReorganizer.java
@@ -73,58 +73,77 @@ public final class BlockCipherReorganizer {
                 @Nullable @Override
                 public List<INode> applyReorganization(
                         @Nonnull INode node, @Nonnull INode parent, @Nonnull List<INode> roots) {
-                    if (!(node instanceof Algorithm algA)) {
+                    if (!(node instanceof Algorithm)) {
                         return roots;
                     }
 
                     List<INode> newRoots = new ArrayList<>(roots);
-                    INode toRemove = null;
-                    INode toKeep = null;
+                    INode currentBase = node;
+                    boolean merged;
+                    do {
+                        merged = false;
+                        INode toRemove = null;
+                        INode toKeep = null;
 
-                    for (INode other : newRoots) {
-                        if (other != node && other instanceof Algorithm algB) {
+                        for (INode other : newRoots) {
+                            if (other != currentBase && other instanceof Algorithm algB) {
+                                Algorithm baseAlg = (Algorithm) currentBase;
 
-                            if (algA.getDetectionContext().equals(algB.getDetectionContext())
-                                    && isRelatedCipherFamily(algA.getName(), algB.getName())) {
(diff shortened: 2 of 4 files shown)


somiljain2006:

san-zrl (commented):
...

san-zrl (changes requested):
### Potential Edge Cases NOT Covered:
- Multiple overlapping roots: If 3+ algorithms overlap (e.g., AES, AES-128, AES-256 all detected), only one pair is processed per iteration. The reorganizer may need mu
- Case sensitivity edge case: equalsIgnoreCase("Unknown") is used, but algorithm name comparison in isRelatedCipherFamily() uses case-sensitive equals(). This could cause issues if algorithm names have inconsistent casing.
- Ambiguous prefix matching: "RC2" and "RC256" would be considered related (RC2 is prefix of RC256 followed by digit '5'), which may not be intended.
- Child key conflicts: When merging children with toKeep.put(child), if both nodes have children of the same Kind, only one survives (the check !toKeep.getChildren().containsKey(child.getKind()) prevents overwriting, but the removed node's child is lost).
- Empty name strings: While null is handled, empty strings ("") are not explicitly checked and could cause unexpected behavior in prefix matching.
- Non-Algorithm BlockCipher nodes: The rule checks instanceof Algorithm but is registered for BlockCipher.class. If BlockCipher nodes exist that aren't Algorithms, they won't be deduplicated.

### Recommendation
The implementation works correctly for the intended use case (BouncyCastle AES + CTR mode detection). However, consider:

- Adding tests for multiple overlapping roots
- Handling child merge conflicts more explicitly
- Adding validation for edge cases in algorithm name matching
- Documenting the reorganizer's behavior when multiple passes are needed

somiljain2006:
@san-zrl Thanks for the review. I've addressed the reported edge cases:

* **Multiple overlapping roots**: deduplication now runs iteratively until no more merges are possible, with test coverage for 3-node overlaps.
* **Case sensitivity**: names are normalized before comparison.
* **Ambiguous prefix matching**: The existing matching logic already requires a non-alphanumeric separator (which prevents cases like RC2 and RC256 from merging), so I have added a regression test to explicitly prove this behavior.
* **Empty names**: handled explicitly with test coverage.

For the remaining points:

* **Child merge conflicts**: existing children on the retained node are preserved, and only non-conflicting children are merged. This behavior is documented in the code.
* **Non-Algorithm BlockCipher nodes**: the affected Bouncy Castle mappings are translated into `Algorithm` nodes, so restricting the scope to Algorithm is safe for this context.

san-zrl (approved):
LGTM. Thanks a lot.

PR #468, opened 2026-06-12
Enable DuplicateDependingFindingsTest

somiljain2006:
This PR enables `DuplicateDependingFindingsTest`, which was previously disabled due to duplicate depending detection rules being propagated from `OAEPEncoding` to its child `RSAEngine`. The test assertions have been updated to verify that the `init` depending detection rule is only associated with `OAEPEncoding` and not the underlying engine. This adds regression coverage and helps prevent duplicate findings from being reintroduced.

Also fixed spotless errors in PycaDigestContextTranslator.java


somiljain2006:

san-zrl (approved):
Looks good to me. Thank you.

PR #470, opened 2026-06-18
Enable DetectionWithSubRuleTest and report dependent rule findings

somiljain2006:
Recursively report detections from child DetectionStores so findings from withDependingDetectionRules(...) reach CheckVerifier alongside root detections. Update DetectionWithSubRuleTest to cover both Cipher.getInstance and cipher.init detections.


somiljain2006:

PR #495, opened 2026-07-22
Feat/python crypto detection

san-zrl:
## Summary

This PR adds detection support for the **PyCrypto / PyCryptodome** library (`Crypto.*`) and reorganises the existing Python detection rule infrastructure so that both `pyca` and `pycrypto` rules share the same translator pipeline.

The rules also work for `pycryptodomex` which exhibits the exact same API as `pycryptodome`. The only difference is the import path (`Cryptography` instead of `Crypto`).  Detection rules check for both possible import paths in forObjectTypes. Rule duplication was avoided by relaxing the type checking (withMethodParameter uses ANY, parameters detected during translation). Only one duplicate rule.

Here's a result that this PR produces with cbomkit when scanning mastercard python client API (no findings before). 
image

---

### 1 · New PyCrypto / PyCryptodome detection rules

Eight new detection-rule classes cover the full PyCryptodome API surface:

| Rule file | Algorithms covered |
|---|---|
| `PythonCryptoCipher` | AES, 3DES, DES, Blowfish, CAST5, RC2, RC4, Salsa20, ChaCha20, ChaCha20-Poly1305, PKCS1-OAEP, PKCS1v15, HPKE |
| `PythonCryptoHash` | MD2, MD5, SHA-1/2/3 family, BLAKE2b/s, RIPEMD-160, SHA-512/t, TupleHash128, cSHAKE256 |
| `PythonCryptoKDF` | PBKDF1, PBKDF2, scrypt, HKDF, bcrypt |
| `PythonCryptoKeyAgreement` | DH, X25519, X448 |
| `PythonCryptoMac` | HMAC, CMAC |
| `PythonCryptoPublicKey` | RSA, DSA, ECC (with `ECCRules()` for ECDSA dependency), ElGamal |
| `PythonCryptoSignature` | DSS (sign + verify), EdDSA (sign + verify), PKCS1v15 (sign + verify), PSS (sign + verify), ECDSA (sign + verify) |
| `PythonCryptoRandom` | Secure random byte generation |

### 2 · Shared PyCA translators for PyCrypto

`pycrypto` detections are wired into the **existing pyca context translators** rather than duplicating translation logic. Three new translators were added to fill gaps:

- `PycaKeyContextTranslator` — unified key material handling for both libraries
- `PycaPublicKeyContextTranslator` — public-key extraction shared by pyca and pycrypto RSA/DSA/ECC
- `PycaRandomContextTranslator` — random context (new, used by `PythonCryptoRandom`)

### 3 · Code structure reorganisation

All pyca detection rules and their test input files moved into a dedicated `pyca/` sub-package, mirroring the new `pycrypto/` layout:

```
rules/detection/
  pyca/          ← moved from detection/ root
    aead/  asymmetric/  fernet/  hash/  kdf/
    keyagreement/  mac/  padding/  symmetric/  wrapping/
  pycrypto/      ← new
    cipher/  hash/  kdf/  keyagreement/
    mac/  publickey/  random/  signature/
```

Test classes `.java` and test input `.py` files follow the same structure under `src/test/files/rules/detection/pyca/` and `pycrypto/`. 

`PythonDetectionRules.java` and `PythonReorganizerRules.java` updated to include all new rule registrations.

### 4 · Mapper additions

| File | Change |
|---|---|
| `PycaCurveMapper` | New — maps PyCrypto/pyca curve names to canonical `EllipticCurve` models |
| `PycaDigestMapper` | Extended with additional hash algorithms |
| `PycaMacMapper` | Extended with CMAC |
| `PycaCipherMapper` | Extended with PyCrypto cipher modes |
| `SignatureReorganizer` | Extended to handle DSS/EdDSA/PSS/ECDSA reorganization |

### 5· Performance

Added memoization to hot translation paths to avoid redundant re-computation during multi-rule traversals.

---

## Tests

55 new JUnit tests under `python/src/test/java/…/pycrypto/` — one per detection scenario — covering approximately **80 % of the new detection rules**:

- **Cipher** (13): AES, 3DES, DES, Blowfish, CAST5, RC2, RC4, Salsa20, ChaCha20, ChaCha20-Poly1305, PKCS1-OAEP, PKCS1v15, HPKE
- **Hash** (16): MD2, MD5, SHA-1/224/256/384/512, SHA3-224/256/384/512, BLAKE2b, BLAKE2s, RIPEMD-160, TupleHash128, cSHAKE256
- **KDF** (5): PBKDF1, PBKDF2, scrypt, HKDF, bcrypt
- **Key agreement** (5): DH, X25519 private/public, X448 private/public
- **MAC** (2): HMAC, CMAC
- **Public key** (4): RSA, DSA, ECC, ElGamal
- **Signature** (10): DSS sign/verify, EdDSA sign/verify, PKCS1v15 sign/verify, PSS sign/verify, ECDSA sign/verify

41 existing pyca tests updated to reflect the restructured package paths.

---

## Directory structure change

| Before | After | |
|---|---|---|
| `detection/aead/Pyca*.java` | `detection/pyca/aead/Pyca*.java` | moved |
| `detection/asymmetric/Pyca*.java` | `detection/pyca/asymmetric/Pyca*.java` | moved |
| `detection/hash\|kdf\|mac\|padding…` | `detection/pyca/hash\|kdf\|mac\|padding…` | moved |
| `test/files/rules/detection/*.py` | `test/files/rules/detection/pyca/*.py` | moved |
| — | `detection/pycrypto/**/*.java` | new |
| — | `test/files/rules/detection/pycrypto/**/*.py` | new |

---

## Translator additions & changes

| Translator | Status | Note |
|---|---|---|
| `PycaKeyContextTranslator` | new | Unified key handling for pyca + pycrypto |
| `PycaPublicKeyContextTranslator` | new | Shared RSA/DSA/ECC public-key extraction |
| `PycaRandomContextTranslator` | new | Random context, used by `PythonCryptoRandom` |
| `PycaCipherContextTranslator` | updated | Extended for PyCrypto cipher modes |
| `PycaKeyDerivationContextTranslator` | updated | Added PBKDF1, bcrypt support |
| `PycaSignatureContextTranslator` | updated | Added DSS, EdDSA, PSS, ECDSA contexts |
| `PycaPrivateKeyContextTranslator` | updated | Simplified; shared logic moved to `PycaKeyContextTranslator` |
| `PycaDigestContextTranslator` | updated | Minor extension |
| `PycaKeyAgreementContextTranslator` | updated | DH / X25519 / X448 support |
| `PycaMacContextTranslator` | updated | CMAC added |

---


n1ckl0sk0rtge (commented):
## PR Review: Feat/python crypto detection (#495)

**Linked Issues:** none — PR has no `Closes #N` reference
**Requirements Coverage:** feature largely delivered, but 3 of the 8 advertised rule groups don't reach production (see MUST FIX 1, 4, 5, 7)
**Architecture:** CLEAN — no layer violations; rules import no `com.ibm.mapper.*`, all 8 holders follow the memoization contract
**Domain Patterns:** 8 blocking findings (crypto mis-mapping + dead detections)
**Blast Radius:** MEDIUM — the `engine` prefix-matching change is provably inert for Java/Go, but `SignatureReorganizer` sits on live Java **and** Go paths with a semantic change and zero Java/Go tests
**Security:** no secrets in fixtures, no new dependencies, no analysis-time IO — clean
**Tests:** 55 new tests, all green (`mvn test -pl python -am` → BUILD SUCCESS, 106 python tests, 0 failures); assertions genuinely exhaustive
**Breaking Changes:** EXTERNAL — `MethodMatcher` public constructor replaced; existing pyca curve detection regresses

### MUST FIX (Blocking)

**1. [Architecture] `PythonCryptoPublicKey.rules()` is never registered — all public-key detection is dead in production**
`python/src/main/java/com/ibm/plugin/rules/detection/PythonDetectionRules.java:81-87`
Only 7 of the 8 pycrypto holders are in `buildRules()`. `RSA.generate()`, `DSA.generate()`, `ECC.generate()`, `ECC.import_key()`, `ElGamal.generate()` produce **zero findings in a real scan**. The 4 publickey tests pass only because they instantiate `PythonCryptoPublicKey.rules()` directly — they give false confidence.
*Fix:* add `PythonCryptoPublicKey.rules().stream(),` to the `Stream.of(...)`.

**2. [Domain] `PycaCurveMapper` drops `SECP521R1` and `SECP256K1` — regression for existing pyca users**
`mapper/src/main/java/com/ibm/mapper/mapper/pyca/PycaCurveMapper.java:70`
`PycaPrivateKeyContextTranslator` previously handled both inline; it now delegates to this mapper, which has neither. Instead it has `case "SECP512R1", "PRIME512V1", "P-512", "P512", "NIST P-512" -> new Secp521r1(...)` — **secp512r1 does not exist**. So `ec.SECP521R1()` and `ec.SECP256K1()` in already-scanned Python code silently stop emitting their curve component.
*Fix:* rename the case set to `"SECP521R1", "PRIME521V1", "P-521", "P521", "NIST P-521"` and re-add `case "SECP256K1" -> new Secp256k1(...)`.

**3. [Domain] Ed25519/Ed448 conflated with X25519/X448 curves**
`mapper/src/main/java/com/ibm/mapper/mapper/pyca/PycaCurveMapper.java:72-75` — `case "ED25519", "CURVE25519" -> Curve25519`. EdDSA signature curves and Montgomery ECDH curves are distinct, and the repo already models them separately (`Ed25519`/`Ed448` in `PycaPrivateKeyContextTranslator:97-98`). `ECC.generate(curve='Ed25519')` is now reported as an ECDH curve.
*Fix:* split the aliases. Note `ECCTestFile.py:4` bakes the conflation into an assertion, so that marker needs updating too.

**4. [Domain] `bcrypt` is detected but never translated — dead detection**
`python/src/main/java/com/ibm/plugin/rules/detection/pycrypto/kdf/PythonCryptoKDF.java:280,293` emit `kind = "bcrypt"`, but `PycaKeyDerivationContextTranslator` has **zero** bcrypt cases and there is no `Bcrypt` model class. `BcryptTestFile.py:6` admits it — `# detected but not mapped` — and `BcryptTest.java:73` asserts `nodes` is empty. The test locks in the gap rather than covering the feature.
*Fix:* add a `Bcrypt` algorithm model + translator case, or drop `BCRYPT`/`BCRYPT_CHECK` until it exists.

**5. [Domain] EdDSA import rules emit `"ECC"`, which no translator handles — dead detection**
`python/src/main/java/com/ibm/plugin/rules/detection/pycrypto/publickey/PythonCryptoPublicKey.java:217,230` build contexts with `algorithm = "ECC"`, but `PycaPublicKeyContextTranslator` handles only `DH|RSA|DSA|ELGAMAL` and `PycaPrivateKeyContextTranslator` handles `EC`, not `ECC`. Both `eddsa.import_private_key` / `import_public_key` rules never translate.
*Fix:* use `"ED25519"`/`"ED448"`, or add an `ECC` case.

**6. [Domain] `"shuflle"` typo**
`python/src/main/java/com/ibm/plugin/rules/detection/pycrypto/random/PythonCryptoRandom.java:58` — `Crypto.Random.random.shuffle()` is never detected. No test exists for `PythonCryptoRandom` at all, which is why it wasn't caught.

**7. [Domain] The four common PBKDF2 forms are commented out of `buildRules()`**
`python/src/main/java/com/ibm/plugin/rules/detection/pycrypto/kdf/PythonCryptoKDF.java:322-325` — only the 5-arg `PBKDF2_WITH_COUNT_AND_HASH` is active, so `PBKDF2(password, salt, 32)` and `PBKDF2(password, salt, 32, count=100000)` are undetected. The private constants remain in the file (lines 149-222).
*Fix:* enable or delete. Same pattern at `PythonCryptoCipher.java:311` (references `PKCS1_OAEP_NEW_WITH_HASH`, a constant that does not exist) and `PythonCryptoPublicKey.java:76-88,118-130,194-206`.

**8. [Breaking Change] `MethodMatcher` public constructor replaced, not overloaded**
`engine/src/main/java/com/ibm/engine/detection/MethodMatcher.java:72`
`MethodMatcher(String[], String[], List<String>)` was mutated in place into a 4-arg `prefixMatch` form. `docs/LANGUAGE_SUPPORT.md:459` points extension authors at this class — this is a source- and binary-incompatible removal for out-of-tree consumers. Nothing in-repo breaks.
*Fix:* keep the 3-arg array constructor delegating `this(..., false)`.

### SHOULD FIX (Non-blocking)

**9. [Domain] All 14 KDF rules use `.inBundle(() -> "PyCa")` instead of `"PyCrypto"`** — `PythonCryptoKDF.java` (14/14; every other pycrypto class correctly uses `"PyCrypto"`). PyCryptodome KDF findings will be attributed to the pyca library in the CBOM.

**10. [Blast Radius] `SignatureReorganizer` mutates a live, aliasable children map** — `mapper/src/main/java/com/ibm/mapper/reorganizer/rules/SignatureReorganizer.java:200-201,263-264`. The `ConcurrentModificationException` fix (`values().forEach(n::put); getChildren().clear();`) iterates and clears the live internal map — and `Algorithm`'s copy constructor **aliases** it (`Algorithm.java:39: this.children = algorithm.getChildren()`), so `clear()` can wipe a different node's children. Both methods are on live **Java** (`JavaReorganizerRules:58`) and **Go** (`GoReorganizerRules:64,66,68`) paths, and this PR adds zero Java/Go tests.
*Fix:* iterate a defensive copy; add a multi-child regression test for Java and Go.

**11. [Scope] `mapper/ciphersuites.json` + `JsonCipherSuites.java` are unrelated build artifacts** — `mapper/download-cipher-suites.sh` curls the file from `ciphersuite.info` and `creatCipherSuiteClass.py` regenerates the Java. The semantic diff is clean (348 → 351 entries: 3 ASCON suites added, 0 removed, 0 modified), but this has nothing to do with Python crypto detection.
*Fix:* `git checkout main -- mapper/ciphersuites.json mapper/src/main/java/com/ibm/mapper/mapper/ssl/json/JsonCipherSuites.java`; land the ASCON suites separately.

**12. [Domain] `DES_3` is fully subsumed by `DES_2` → duplicate findings** — `PythonCryptoCipher.java:99-112` vs `:84-97`. Since `prefixMatchesParameters` accepts `actual >= expected`, `DES.new(key, DES.MODE_CBC, iv)` matches both rules. *Fix:* delete `DES_3`.

**13. [Domain] ChaCha20 / ChaCha20-Poly1305 get no Encrypt/Decrypt nodes** — `PythonCryptoCipher.java:186-208` end with `.withoutDependingDetectionRules()` while every sibling attaches `List.of(ENCRYPT, DECRYPT)`.

**14. [Domain] `pkcs1_15` signatures lose their PKCS#1 v1.5 padding** — `PythonCryptoSignature.java:81` → `ValueActionFactory("RSA")`, indistinguishable from a raw RSA signature. Compare `pss` → `"RSA-PSS"` → `RSAssaPSS`. The padding branch at `PycaSignatureContextTranslator.java:82` is still `case "PKCS1V15" -> Optional.empty(); // TODO`.

**15. [Domain] `DSS` rules lack `withOtherParameters()`** — `PythonCryptoSignature.java:123-152` require exactly 2 params, so the documented `DSS.new(key, 'fips-186-3', encoding='der')` is missed. The ECDSA twins do use it. (DSS-vs-ECDSA discrimination itself is correct — it keys off the first parameter's type, so there is no double counting.)

**16. [Domain] HMAC/CMAC bind the algorithm to the 2nd positional parameter, which is `msg`** — `PythonCryptoMac.java:42-64`. The real signatures are `HMAC.new(key, msg=b'', digestmod=None)` and `CMAC.new(key, msg=b'', ciphermod=None)`. `HMAC.new(key, data, SHA256)` is missed.

**17. [Domain] Poly1305 / KMAC128 / KMAC256 registered as digests, not MACs** — `PythonCryptoHash.java:41-68`. This contradicts `PythonCryptoMac.java:66-78`, where the Poly1305 MAC rule was deliberately disabled.

**18. [Domain] All hash rules use `.withAnyParameters()`, dropping size/variant** — `PythonCryptoHash.java:83-92`. `SHA512.new(truncate='256')` reports plain SHA-512 even though `PycaDigestMapper` already has `SHA512_224`/`SHA512_256` ready; BLAKE2b/s lose their digest size.

**19. [Domain] The HPKE rule likely cannot fire** — `PythonCryptoCipher.java:269-285` uses two positional parameters, but PyCryptodome's `HPKE.new()` is keyword-only. `HPKETestFile.py` uses invented syntax (`from Crypto.Protocol.HPKE.AEAD import AES128_GCM` — `AEAD` is a class, not a submodule), so the green test does not prove real-world matching. Worth confirming against a real install.

**20. [Tests] Three new components have no test at all** — `PythonCryptoRandom` (both rules), `PycaRandomContextTranslator` (0% coverage; only reachable from it), and `PycaCurveMapper` (20 branches, no test anywhere in `mapper/src/test/`). The missing curve test is exactly what would have caught MUST FIX 2.

**21. [Tests] The `Cryptodome.*` branch is structurally untested** — zero of the 55 new `.py` files use the `Cryptodome` prefix. `DSS_CRYPTODOME` and `ECDSA_CRYPTODOME` exist solely for it and are never exercised. Given the PR's headline claim of pycryptodomex support, this is the largest coverage gap.

**22. [Tests] The new core-engine matching mode has no engine test** — `withOtherParameters()` / `SOME_WITH_REMAINDER` / `prefixMatchesParameters` appear nowhere under any `src/test/`. Used 15 times, verified only indirectly through Python integration tests.

**23. [Architecture] Unchecked `(PublicKeyEncryption)` casts on `Optional<? extends Algorithm>`** — `PycaKeyContextTranslator.java:79`, `PycaPublicKeyContextTranslator.java:79`, `PycaPrivateKeyContextTranslator.java:74`. Safe only by accident today; a `ClassCastException` here aborts the Sonar analysis of the file. *Fix:* narrow `PycaCurveMapper.parse` to `Optional<EllipticCurveAlgorithm>`.

**24. [Architecture] The key-algorithm switch is duplicated across three translators and is already drifting** — `ELGAMAL` was added to two of them, `FERNET`/`EC_IMPORT` to only one. *Fix:* extract an `IMapper` into `mapper/.../mapper/pyca/`.

**25. [Docs] `withOtherParameters()` is absent from the documented builder grammar** — `docs/DETECTION_RULE_STRUCTURE.md:172,178`. Third parties following this doc will not discover the new DSL step.

**26. [Breaking Change] `MethodMatcherSerializer` does not emit `prefixMatch`** — exported rule JSON cannot distinguish prefix from exact arity, misrepresenting all 15 new rules.

**27. [Domain] `PycaSecretContextTranslator` is now dead code** — unwired from `PythonTranslator` (replaced by `PycaKeyContextTranslator` for `KeyContext`) but still present. Confirm no pyca regression, then delete.

**28. [Domain] bcrypt `cost` is log2 of rounds, mapped straight to `ITERATIONS`** — `PythonCryptoKDF.java:269-282` reports `cost=10` as 10 iterations instead of 1024.

**29. [Docs] Broken README link** — `README.md:39`: `[PyCryptodome(x)][https://www.pycryptodome.org/]` is reference-link syntax with no defined reference, so GitHub renders the literal text. Use `(...)` instead. Also revert the trailing-whitespace-only change on line 38.

### SUGGESTIONS

30. `scrypt`'s `r` is a block-size *factor* (block = 128·r bytes), reported as raw bytes — `PythonCryptoKDF.java:258-261`
31. Append `SOME_WITH_REMAINDER` at the end of the enum rather than inserting it — safe today (no `ordinal()`/serialization use anywhere), defensive convention only
32. Guard `prefixMatchesParameters` against an empty `expectedTypes` list (it would match every invocation) — currently unreachable, but one refactor away
33. `PycaCipherContextTranslator.java:109-122` compares raw `asString()` while neighbouring switches use `.toUpperCase().trim()`
34. `Crypto.Protocol.DH.key_agreement` is hard-coded to `"ECDH"` even though the X25519/X448 import rules feed it
35. `PycaCurveMapper` has phantom aliases (`PRIME192V1`/`PRIME224V1`/`PRIME384V1` are not real X9.62 names) and partial brainpool coverage — suggests the list was extrapolated rather than taken from PyCryptodome's curve table
36. Delete the ~20 lines of commented-out original code in `SignatureReorganizer.java:194-205,257-268`

### What's good here

Worth calling out explicitly — several things were done carefully:

- **The 95-file rename is flawless.** Zero stale FQN references repo-wide, including `rules/`, plugin resources, and rule metadata; all 96 `PythonCheckVerifier.verify(...)` path strings resolve. The pyca move is a pure rename with 0 changed lines in the rule bodies.
- **The memoization contract is followed perfectly** across all 8 new classes, including all four suppliers in `PythonCryptoPublicKey`.
- **Layering is respected** — no rule class imports `com.ibm.mapper.*`; `PycaCurveMapper` is correctly language-agnostic.
- **Translator dispatch ordering is correct and safe** — `IDetectionContext.is()` uses exact class equality rather than `isAssignableFrom`, and the new `KeyContext` branch was placed after the subtype branches anyway.
- **Test assertions are genuinely exhaustive** (12–46 assertions per class, covering both the detection store and the translated nodes) — no stubs. `BcryptTest`'s empty-node assertion is honest documentation of a gap, not a cheat.
- **The broad `ENCRYPT`/`DECRYPT`/`SIGN`/`VERIFY` rules are correctly excluded from top-level registration**, so there is no repo-wide match on every `.encrypt()` call.
- **`prefixMatch` is provably inert for Java/Go** — gated behind the new opt-in builder step and used only in `pycrypto/`.

---

**Bottom line:** the reorganization and test discipline are solid, but the 8 blocking issues mean the feature does not fully work as advertised — one entire rule class is unregistered, three rule groups are dead detections, and the shared curve mapper regresses existing pyca output. The `engine` and `mapper/ciphersuites.json` changes would be better split into separate PRs.

---
> This review was generated by AI (Claude). Findings may contain errors — please verify before acting on them.

san-zrl:
Hi @n1ckl0sk0rtge,

Thanks for the review. I modified the code in response and pushed my changed .
Note that I rebased the branch and organized the commit history into logical, grouped steps.

Here's what I did:

### Must Fix

**M1**: PythonCryptoPublicKey.rules() was redesigned — it now only contains ElGamal. The are marked as obsolete keys and should be detected stand-alone. A class-level comment documents this.

RSA_GENERATE, DSA_GENERATE, ECC_GENERATE, etc. are intentionally not top-level rules. They are reached as dependent rules from PythonCryptoSignature — for example, PKCS1V15 and PSS call PythonCryptoPublicKey.RSARules(), DSS call PythonCryptoPublicKey.DSARules(), and ECDSA, EDDSA and DH_KEY_AGREEMENT call PythonCryptoPublicKey.ECCRules(). 

**M2**: The typo — SECP512R1 → SECP521R1 (and all its aliases) — fixed
1. The original broken case replaced with "SECP521R1", "PRIME521V1", "P-521", "P521", "NIST P-521" 
All five label strings now correctly identify the 521-bit NIST prime curve.
2. The missing SECP256K1 case — re-added

**M3**: Fixed. 
Concern | Status
-- | --
ED25519 mapped to Curve25519 (wrong) | Fixed — now maps to Edwards25519
CURVE25519 still maps to Curve25519 | Correct — preserved
Same split applied to 448-bit family | Done — ED448 → Edwards448, CURVE448 → Curve448
ECCTestFile.py marker updated | Done — EC-Edwards25519
ECCTest.java assertion updated | Done — asserts Edwards25519 throughout. If the mapper still conflated ED25519 → Curve25519, these assertions would fail.

**M4**: Fixed. Removed bcrypt rule and test for the moment.
Bcrypt support could be an issue on its own. Support exists in BouncyCastle (not detected) and pycryptodome.
For python (complementing python cryptography) and for golang there are separate bcrypt libs.  Our mapper does
not provide a BCrypt algorithm, so there's currently no translation target.

**M5**: Fixed. Both detection rules now emit "EC" instead of "ECC" so that detections produce translated output. Tests
could remain unchanged. The suggested modification to emit Ed25519/Ed448 keys does not work since the EDDSA_IMPORT_PUBLIC/PRIVATE_KEY rules cannot infer the precise key type from the call site alone. Therefore the rules produce Elliptic Curve keys which is the correct generalization.

**M6**: Fixed the typo.

**M7**: Fxed. All five PBKDF2 variants are now active in buildRules(): PBKDF2, PBKDF2_WITH_HASH,
PBKDF2_WITH_COUNT, PBKDF2_WITH_HASH_AND_COUNT, PBKDF2_WITH_COUNT_AND_HASH
Non-existent PKCS1_OAEP_NEW_WITH_HASH removed.

**M8**: Fixed as prescribed.

### SHOULD FIX

**S9**: All KDF rules use now .inBundle(() -> "PyCrypto") as suggested.

**S10**: Fixed, but no multi-child regression tests for Java and Go.
1. Both moveNodesFromUnderFunctionalityUnderNode and moveNodesFromUnderFunctionalityUnderParent now snapshot the children into an ArrayList before iterating, and remove each entry individually with removeChildOfType rather than bulk clear(). This eliminates both the ConcurrentModificationException and the aliased-map corruption in a single change.
2.  Root-cause fix: The copy constructors at Algorithm.java:39 and Key.java:48 now both do new HashMap<>(...) instead of assigning the live map reference directly. Any future code that mutates a re-kind copy no longer corrupts the source node's children.

**S11**: No change

**S12**: DES_3 rule removed. Update of Cipher rules AES, DES, DES3, BLOWFISH, CAST, ARC2.

**S13**: Fixed. Both rules now end with .withDependingDetectionRules(List.of(ENCRYPT, DECRYPT)), consistent with all other cipher rules.

**S14**: Fixed.
1. Detection rule — "RSA" → "RSA-PKCS1V15".
2. PycaSignatureContextTranslator now handles "RSA-PKCS1V15" in the main ValueAction switch.

**S15**: Fixed. Both DSS (line 140) and DSS_CRYPTODOME (line 156) now have .withOtherParameters() after the typed key parameter, with a comment explicitly listing what the extra args cover. This means DSS.new(key, 'fips-186-3'), DSS.new(key, 'fips-186-3', encoding='der'), and any other arity variant all match. The fix is symmetric with ECDSA (line 172) and ECDSA_CRYPTODOME (line 188), which already had it.

**S16**: Fixed. The original single-rule approach — which bound the algorithm to the 2nd positional parameter and thus confused msg with the hash/cipher module — was replaced with a two-rule pattern for both HMAC and CMAC.

**S17**: "KMAC128", "KMAC256", and "Poly1305" were wrongly detected as hashes and emitted as if they were digests. 
"KMAC128", "KMAC256", and "Poly1305" were moved to PythonCryptoMac.java which correctly detects them under MacContext. PycaMacContextTranslator.java routes them to PycaMacMapper.java which correctly maps them as MACs. Tests for "KMAC128", "KMAC256", and "Poly1305" were added to python/src/test/java/com/ibm/plugin/rules/detection/pycrypto/mac.

**S18**: Not addressed, accepted as a limitation.
1. SHA512 allows two named parameters "data" and "truncate", both of type str. When using detection rules with positional heuristics there's no way to distinguish the parameters.
2. BLAKE2s/BLAKE2b allow digest size to be passed as bit or byte values via different named parameters. Number of
detection rules would explode.

**S19**: Removed attempted detection of HPKE and corresponding test. The case is similar to M4 (Bcrypt).
In our mapper model there is no HPKE algorithm and thus no translation target for a HPKE detection rule.
I guess this is also the reason as to why  our pyca detection rules don't detect HPKE either even though
the pyca library implements it.

**S20, S21, S22**: Missing tests not addressed.

**S23**: Narrowed PycaCurveMapper.parse to Optional<EllipticCurveAlgorithm>, cast exception now impossible. Removed cast in PycaKeyContextTranslator. The other two casts are necessary. 

**S24**: Fixed. Created PycaKeyBasedAlgorithmMapper and wired it into PycaKeyContextTranslator, PycaPrivateKeyContextTranslator and PycaPublicKeyContextTranslator replacing the drifting switch statements.

**S25**: Fixed. Added withOtherParameters() to the grammar and explained it in the text below.

**S26**: Fixed.
1. MethodMatcher — stored the prefixMatch flag as a proper field across all constructors and exposed it via isPrefixMatch().
2. MethodMatcherSerializer — emits "prefixMatch": true/false into the exported rule JSON so readers can distinguish prefix-arity from exact-arity matchers.
3. DetectionRuleStore — appends the prefixMatch flag to the matcher ID string, preventing two rules that differ only in arity mode from being silently deduplicated during graph export.

**S27**: Fixed. Removed dead PycaSecretContextTranslator..java

**S28**: Void. Bcrypt removed in response to M4.

**S29**. Fixed broken link and trailing whitespace in README.md.

### Suggestions

**S30:** Left as is
**S31:** Left as is
**S32**: Added !expectedTypes.isEmpty() as guard to prefixMatchesParameters()
**S33**: Fixed. Use toUpperCase().trim()
**S34**: Update detection rules in PyhonCryptoKeyAgreement.java that generate ECDH detections with ECC keys.
Added two reorganiser rules that convert the ECDH to X25519 and X448 if the curves in the ECC keys are 25519 or 448.
**S35**: The PRIME curve names were there before. Left as is
**S36**: Deleted commented out code.

python/src/main/java/com/ibm/plugin/rules/detection/pycrypto/kdf/PythonCryptoKDF.java
```diff
@@ -0,0 +1,326 @@
...
+                    .asChildOfParameterWithId(-1)
+                    .withMethodParameter("int") // count
+                    .shouldBeDetectedAs(new IterationCountFactory<>())
+                    .asChildOfParameterWithId(-1)
+                    .buildForContext(
+                            new KeyDerivationFunctionContext(Map.of("kind", "pycrypto-pbkdf1")))
+                    .inBundle(() -> "PyCrypto")
+                    .withoutDependingDetectionRules();
+
+    private static final IDetectionRule<Tree> PBKDF1_WITH_HASH =
```
n1ckl0sk0rtge: **MUST FIX — this rule overlaps `PBKDF1_WITH_COUNT` and produces duplicate CBOM components.**

`PBKDF1_WITH_COUNT` (:64) declares `(ANY, ANY, int, int)`; this rule declares `(ANY, ANY, int, ANY)`. `MethodMatcher.substituteAny` turns `ANY` into an always-true predicate, and `PythonBaseDetectionRule` iterates **all** rules per call expression with no first-match break. So the canonical call

```python
PBKDF1(password, salt, 16, 1000)
```

matches **both** rules: one emits the correct `NumberOfIterations`, the other feeds the int literal `1000` into `AlgorithmFactory`. One call site, two `PBKDF1` roots in the CBOM.

Separately, the modelling here does not match the library. The real signature is:

```python
PBKDF1(password, salt, dkLen, count=1000, hashAlgo=None)
```

`hashAlgo` is the **5th** parameter and keyword-only in practice — it is not the 4th positional. `PBKDF1_WITH_HASH_AND_COUNT` (:124) models the ordering `(dkLen, hashAlgo, count)`, which does not exist in PyCryptodome at all.

To make the pair disjoint today, the hash parameter has to be typed rather than `ANY`. Because `withMethodParameter` takes a single type string and `resolveFullyQualifiedNameStringType` only does prefix wildcards, `"Crypto.Hash.*"` will not match `Cryptodome.Hash.SHA256` — so this needs the same per-root split already used for `DSS`/`DSS_CRYPTODOME`. That is the cost of closing it without an engine change; the `withNamedMethodParameter` proposal removes the need for the split entirely.

python/src/main/java/com/ibm/plugin/rules/detection/pycrypto/kdf/PythonCryptoKDF.java
```diff
@@ -0,0 +1,326 @@
...
+                    .forObjectTypes("Crypto.Protocol.KDF", "Cryptodome.Protocol.KDF")
+                    .forMethods("PBKDF2")
+                    .shouldBeDetectedAs(new ValueActionFactory<>("PBKDF2"))
+                    .withMethodParameter(ANY) // password
+                    .withMethodParameter(ANY) // salt
+                    .withMethodParameter("int") // dkLen
+                    .shouldBeDetectedAs(new KeySizeFactory<>(Size.UnitType.BYTE))
+                    .asChildOfParameterWithId(-1)
+                    .withMethodParameter("int") // count
+                    .shouldBeDetectedAs(new IterationCountFactory<>())
```
n1ckl0sk0rtge: **MUST FIX — the iteration count is silently detached from the PBKDF2 node.**

Every sibling rule follows `IterationCountFactory` with `.asChildOfParameterWithId(-1)` (:114, :214, :236). Without it `DetectableParameter.getShouldBeMovedUnder()` is empty, so `DetectionStore:320` takes the `else` branch and records the count as a *sibling value* rather than a child of the KDF.

Net effect: `PBKDF2(password, salt, 32, 1)` emits a PBKDF2 component with **no iteration count** — a 1-iteration PBKDF2 is indistinguishable from a 600 000-iteration one in the CBOM. `PBKDF2TestFile.py` only exercises the 5-arg `count=`/`hmac_hash_module=` form, which routes to `PBKDF2_WITH_COUNT_AND_HASH`, so this is invisible to the suite.

```suggestion
                    .shouldBeDetectedAs(new IterationCountFactory<>())
                    .asChildOfParameterWithId(-1)
```

python/src/main/java/com/ibm/plugin/rules/detection/pycrypto/kdf/PythonCryptoKDF.java
```diff
@@ -0,0 +1,326 @@
...
+                    .buildForContext(
+                            new KeyDerivationFunctionContext(Map.of("kind", "pycrypto-pbkdf2")))
+                    .inBundle(() -> "PyCrypto")
+                    .withoutDependingDetectionRules();
+
+    private static final IDetectionRule<Tree> PBKDF2_WITH_HASH =
```
n1ckl0sk0rtge: **MUST FIX — same `ANY`-vs-`int` overlap as the PBKDF1 pair above.**

`PBKDF2_WITH_COUNT` (:163) declares `(ANY, ANY, int, int)`, this rule declares `(ANY, ANY, int, ANY)`. `PBKDF2(password, salt, 32, 100000)` matches both, yielding duplicate PBKDF2 components for one call site.

python/src/main/java/com/ibm/plugin/rules/detection/pycrypto/kdf/PythonCryptoKDF.java
```diff
@@ -0,0 +1,326 @@
...
+                    .forMethods("HKDF")
+                    .shouldBeDetectedAs(new ValueActionFactory<>("HKDF"))
+                    .withMethodParameter(ANY) // master
+                    .withMethodParameter("int") // keylen
+                    .shouldBeDetectedAs(new KeySizeFactory<>(Size.UnitType.BYTE))
+                    .asChildOfParameterWithId(-1)
+                    .withMethodParameter(ANY) // salt
+                    .withMethodParameter(ANY) // Crypto.Hash.* or Cryptodome.Hash.* (hash_mod)
+                    .shouldBeDetectedAs(new AlgorithmFactory<>())
+                    .asChildOfParameterWithId(-1)
```
n1ckl0sk0rtge: **SHOULD FIX — `HKDF` calls passing `num_keys` or `context` are not detected.**

The real signature is `HKDF(master, key_len, salt, hashmod, num_keys=1, context=None)`. This rule declares exactly 4 parameters and, unlike `SCRYPT` (:259) and `SP800_108_COUNTER` (:295), has no `.withOtherParameters()` — so exact arity matching rejects any call that passes the optional arguments.

```suggestion
                    .asChildOfParameterWithId(-1)
                    .withOtherParameters() // num_keys, context
```

python/src/main/java/com/ibm/plugin/rules/detection/pycrypto/cipher/PythonCryptoCipher.java
```diff
@@ -0,0 +1,258 @@
...
+                    .inBundle(() -> "PyCrypto")
+                    .withoutDependingDetectionRules();
+
+    private static final IDetectionRule<Tree> DECRYPT =
+            new DetectionRuleBuilder<Tree>()
+                    .createDetectionRule()
+                    .forObjectTypes(ANY)
+                    .forMethods("decrypt", "decrypt_and_verify")
+                    .shouldBeDetectedAs(new CipherActionFactory<>(CipherAction.Action.DECRYPT))
+                    .withMethodParameter(ANY)
```
n1ckl0sk0rtge: **MUST FIX — this blocks every AEAD and PKCS#1 v1.5 decrypt detection.**

`decrypt_and_verify` is listed in `forMethods`, but a single `.withMethodParameter(ANY)` means exact arity 1. The real signatures take more:

- `decrypt_and_verify(ciphertext, received_mac_tag)` — 2 args
- `PKCS1_v1_5.decrypt(ciphertext, sentinel[, expected_pt_len])` — 2–3 args

So every GCM / EAX / CCM / OCB / SIV and PKCS1v15 decrypt goes undetected, and those ciphers land in the CBOM with an **Encrypt-only** functionality set. No fixture exercises `decrypt_and_verify`, which is why the suite stays green.

```suggestion
                    .withMethodParameter(ANY)
                    .withOtherParameters()
```

python/src/main/java/com/ibm/plugin/rules/detection/pycrypto/cipher/PythonCryptoCipher.java
```diff
@@ -0,0 +1,258 @@
...
+        // private
+    }
+
+    private static final IDetectionRule<Tree> ENCRYPT =
+            new DetectionRuleBuilder<Tree>()
+                    .createDetectionRule()
+                    .forObjectTypes(ANY)
+                    .forMethods("encrypt", "encrypt_and_digest")
+                    .shouldBeDetectedAs(new CipherActionFactory<>(CipherAction.Action.ENCRYPT))
+                    .withMethodParameter(ANY)
```
n1ckl0sk0rtge: **SHOULD FIX — same arity issue on the encrypt side.**

`encrypt_and_digest(plaintext)` is 1 arg so it matches, but `encrypt(plaintext, output=...)` and `encrypt_and_digest(plaintext, output=...)` are not detected.

```suggestion
                    .withMethodParameter(ANY)
                    .withOtherParameters()
```

python/src/main/java/com/ibm/plugin/rules/detection/pycrypto/random/PythonCryptoRandom.java
```diff
@@ -0,0 +1,77 @@
...
+                    .withoutDependingDetectionRules();
+
+    private static final IDetectionRule<Tree> RANDOM_FUNC =
+            new DetectionRuleBuilder<Tree>()
+                    .createDetectionRule()
+                    .forObjectTypes("Crypto.Random.random", "Cryptodome.Random.random")
+                    .forMethods(
+                            "getrandbits", "randrange", "randint", "choice", "shuffle", "sample")
+                    .shouldBeDetectedAs(new ValueActionFactory<>("PRNG"))
+                    .withMethodParameter(ANY)
```
n1ckl0sk0rtge: **MUST FIX — most of the listed methods can never match.**

Exact arity 1 against the real signatures:

| method | arity | matches? |
|---|---|---|
| `getrandbits(k)` | 1 | yes |
| `choice(seq)` | 1 | yes |
| `shuffle(seq)` | 1 | yes |
| `randrange(start, stop[, step])` | 2–3 | **no** |
| `randint(N, M)` | 2 | **no** |
| `sample(population, k)` | 2 | **no** |

Half the declared surface is dead. Compounding it, there is no `pycrypto/random/` test class and no fixture anywhere (`find python/src/test/files -ipath '*random*'` returns nothing), so this rule and the new `PycaRandomContextTranslator` are entirely unexercised — prior-review S20, still open.

A top-level `ValueActionFactory` is present, so `withAnyParameters()` is also permitted by the documented grammar if you prefer it here.

```suggestion
                    .withMethodParameter(ANY)
                    .withOtherParameters()
```

python/src/main/java/com/ibm/plugin/rules/detection/pycrypto/signature/PythonCryptoSignature.java
```diff
@@ -0,0 +1,256 @@
...
+
+    @Nonnull
+    public static List<IDetectionRule<Tree>> rules() {
+        return RULES.get();
+    }
+
+    @Nonnull
+    private static List<IDetectionRule<Tree>> buildRules() {
+        return List.of(
+                PKCS1V15, PSS, PSS_MGF1, DSS, DSS_CRYPTODOME, ECDSA, ECDSA_CRYPTODOME, EDDSA);
```
n1ckl0sk0rtge: **MUST FIX — `PSS_MGF1` should not be a top-level rule.**

MGF1 is a mask-generation function, not a signature algorithm. Registered top-level it produces a `SignatureContext` root that translates to a standalone `MGF1` node (`PycaSignatureContextTranslator:92`), so a single `pss.new(key, mask_func=lambda x, y: pss.MGF1(x, y, SHA256))` emits a spurious `MGF1` component alongside the real `RSA-PSS`. It also attaches `SIGN`/`VERIFY` depending rules to a function that returns bytes.

The pyca side already models this correctly — `pyca/asymmetric/PycaRSA.java:78` uses MGF1 as a *depending* rule of the signature rule.

The suggestion below is the minimal step; the complete fix is to wire `PSS_MGF1` in as a depending rule of `PSS` and drop its SIGN/VERIFY children (it will otherwise sit unused at :114).

```suggestion
                PKCS1V15, PSS, DSS, DSS_CRYPTODOME, ECDSA, ECDSA_CRYPTODOME, EDDSA);
```

python/src/test/files/rules/detection/pycrypto/kdf/PBKDF1TestFile.py
```diff
@@ -0,0 +1,7 @@
+from Crypto.Protocol.KDF import PBKDF1
+from Crypto.Hash import SHA256
+
+def test_pbkdf1():
+    password = b"password"
+    salt = b"salt1234"
+    key = PBKDF1(password, salt, 16, SHA256) # Noncompliant {{(PasswordBasedKeyDerivationFunction) PBKDF1-SHA-256}}
```
n1ckl0sk0rtge: **MUST FIX — this is not valid PyCryptodome; the fixture was written to fit the rule.**

`PBKDF1(password, salt, dkLen, count=1000, hashAlgo=None)` — the 4th positional is `count`, so this call binds `SHA256` to `count`, not to `hashAlgo`. The expected `PBKDF1-SHA-256` result is therefore unreachable in real code.

This is the same defect class as the invented `PKCS1_OAEP_NEW_WITH_HASH` from the last round: the fixture validates the rule against itself rather than against the library.

```suggestion
    key = PBKDF1(password, salt, 16, hashAlgo=SHA256) # Noncompliant {{(PasswordBasedKeyDerivationFunction) PBKDF1-SHA-256}}
```

python/src/test/files/rules/detection/pycrypto/mac/CMACTestFile.py
```diff
@@ -0,0 +1,5 @@
+from Crypto.Hash import CMAC
+from Crypto.Cipher import AES
+
+key = b'some key'
+cmac = CMAC.new(key, AES) # Noncompliant {{(Mac) CMAC-AES}}
```
n1ckl0sk0rtge: **SHOULD FIX — invalid PyCryptodome call in the fixture.**

`CMAC.new(key, msg=b'', ciphermod=None, ...)` — the 2nd positional is `msg`, not `ciphermod`. As written this passes the `AES` module as the message. The rule still matches (fixture-only issue), but it documents an API that does not exist.

```suggestion
cmac = CMAC.new(key, ciphermod=AES) # Noncompliant {{(Mac) CMAC-AES}}
```

mapper/src/main/java/com/ibm/mapper/mapper/pyca/PycaCurveMapper.java
```diff
@@ -0,0 +1,113 @@
...
+            @Nullable String str, @Nonnull DetectionLocation detectionLocation) {
+        if (str == null) {
+            return Optional.empty();
+        }
+
+        @Nonnull String curve = str;
+        return switch (curve.toUpperCase().trim()) {
+            case "SECP192R1", "PRIME192V1", "P-192", "P192", "NIST P-192" ->
+                    Optional.of(new EllipticCurveAlgorithm(new Secp192r1(detectionLocation)));
+            case "SECP224R1", "PRIME224V1", "P-224", "P224", "NIST P-224" ->
```
n1ckl0sk0rtge: **SHOULD FIX — `PRIME224V1` is not a real curve name.**

ANSI X9.62 defines only `prime192v1..v3`, `prime239v1..v3` and `prime256v1`. There is no `prime224v1`, and neither PyCryptodome nor pyca/cryptography accepts it — this alias can never fire.

Note this partly rebuts the "the PRIME curve names were there before" response to S35: `PRIME192V1` (:64) and `PRIME256V1` (:68) are legitimate and pre-existing, but `PRIME224V1`, `PRIME384V1` and `PRIME521V1` are new here and invalid.

```suggestion
            case "SECP224R1", "P-224", "P224", "NIST P-224" ->
```

mapper/src/main/java/com/ibm/mapper/mapper/pyca/PycaCurveMapper.java
```diff
@@ -0,0 +1,113 @@
...
+                    Optional.of(new EllipticCurveAlgorithm(new Secp224r1(detectionLocation)));
+            case "SECP256R1", "PRIME256V1", "P-256", "P256", "NIST P-256" ->
+                    Optional.of(new EllipticCurveAlgorithm(new Secp256r1(detectionLocation)));
+            case "SECP384R1", "PRIME384V1", "P-384", "P384", "NIST P-384" ->
```
n1ckl0sk0rtge: **SHOULD FIX — `PRIME384V1` does not exist in X9.62 (see the `PRIME224V1` note above).**

```suggestion
            case "SECP384R1", "P-384", "P384", "NIST P-384" ->
```

mapper/src/main/java/com/ibm/mapper/mapper/pyca/PycaCurveMapper.java
```diff
@@ -0,0 +1,113 @@
...
+                    Optional.of(new EllipticCurveAlgorithm(new Secp384r1(detectionLocation)));
+            case "SECP521R1", "PRIME521V1", "P-521", "P521", "NIST P-521" ->
```
n1ckl0sk0rtge: **SHOULD FIX — `PRIME521V1` does not exist in X9.62 (see the `PRIME224V1` note above).**

```suggestion
            case "SECP521R1", "P-521", "P521", "NIST P-521" ->
```

python/src/main/java/com/ibm/plugin/translation/translator/contexts/PycaCipherContextTranslator.java
```diff
@@ -85,25 +93,49 @@ public final class PycaCipherContextTranslator implements IContextTranslation<Tr
...
         } else if (value instanceof Mode<Tree> mode) {
             return switch (mode.asString().toUpperCase().trim()) {
-                case "CBC" -> Optional.of(new CBC(detectionLocation));
-                case "CTR" -> Optional.of(new CTR(detectionLocation));
-                case "OFB" -> Optional.of(new OFB(detectionLocation));
-                case "CFB" -> Optional.of(new CFB(detectionLocation));
+                case "CBC", "MODE_CBC" -> Optional.of(new CBC(detectionLocation));
+                case "CTR", "MODE_CTR" -> Optional.of(new CTR(detectionLocation));
+                case "OFB", "MODE_OFB" -> Optional.of(new OFB(detectionLocation));
+                case "MODE_OCB" -> Optional.of(new OCB(detectionLocation));
```
n1ckl0sk0rtge: **SHOULD FIX — PyCryptodome's `MODE_OCB` is OCB3 (RFC 7253).**

The pyca AEAD path in this same file maps `AESOCB3` to `new OCB(3, detectionLocation)`; leaving this one unversioned means the same construction is reported two different ways depending on which library was scanned.

```suggestion
                case "MODE_OCB" -> Optional.of(new OCB(3, detectionLocation));
```

python/src/main/java/com/ibm/plugin/translation/translator/contexts/PycaCipherContextTranslator.java
```diff
@@ -85,25 +93,49 @@ public final class PycaCipherContextTranslator implements IContextTranslation<Tr
...
+                    RSA rsaOaep = new RSA(PublicKeyEncryption.class, detectionLocation);
+                    rsaOaep.put(new OAEP(detectionLocation));
+                    yield Optional.of((INode) rsaOaep);
+                }
+                case "PKCS1_V1_5" -> {
+                    RSA rsaPkcs1 = new RSA(PublicKeyEncryption.class, detectionLocation);
+                    rsaPkcs1.put(new PKCS1(detectionLocation));
+                    yield Optional.of((INode) rsaPkcs1);
+                }
+                case "HPKE" -> Optional.of(new EllipticCurveAlgorithm(detectionLocation));
```
n1ckl0sk0rtge: **SHOULD FIX — dead case with an incorrect mapping.**

No detection rule emits `"HPKE"` any more (the HPKE rules were removed in response to S19); `grep -rn '"HPKE"' python/src/main` finds only this line. It is also cryptographically wrong: HPKE (RFC 9180) is a hybrid KEM + AEAD scheme, not an elliptic-curve algorithm.

Worth deleting along with the rules it belonged to, so a future reader does not take it as a supported mapping. Note the PR body still lists HPKE under `PythonCryptoCipher`.

python/src/main/java/com/ibm/plugin/translation/translator/contexts/PycaSignatureContextTranslator.java
```diff
@@ -82,7 +85,16 @@ public final class PycaSignatureContextTranslator implements IContextTranslation
                 };
             } else {
                 return switch (value.asString().toUpperCase().trim()) {
+                    case "RSA" -> Optional.of(new RSA(Signature.class, detectionLocation));
+                    case "DSS" -> Optional.of(new DSS(detectionLocation));
+                    case "ECDSA" -> Optional.of(new ECDSA(detectionLocation));
+                    case "EDDSA" -> Optional.of(new EdDSA(detectionLocation));
                     case "MGF1" -> Optional.of(new MGF1(detectionLocation));
+                    case "RSA-PKCS1V15" -> {
```
n1ckl0sk0rtge: **SHOULD FIX — while you are in this switch: `case "PKCS1V15"` on :83 is a detection with no translation target.**

(Commenting here because :83 is pre-existing and outside the diff.) That case reads:

```java
case "PKCS1V15" -> Optional.empty(); // TODO
```

Meanwhile `pyca/asymmetric/PycaRSA.java:84-96` emits `ValueActionFactory("PKCS1v15")` into a `SignatureContext(kind=padding)` with a comment stating it exists "to trigger the translation". So pyca RSA PKCS#1 v1.5 **signature** padding is detected and then silently dropped.

The governing principle adopted in this PR for removing bcrypt (M4) and HPKE (S19) was that every detection must have a translation target. Same standard applies here, and the model class already exists — the new `RSA-PKCS1V15` arm you added right below is exactly the shape it needs:

```java
case "PKCS1V15" -> Optional.of(new PKCS1(detectionLocation));
```

python/src/test/files/rules/detection/pycrypto/kdf/BcryptTestFile.py
```diff
@@ -0,0 +1,6 @@
+from Crypto.Protocol.KDF import bcrypt
+
+def test_bcrypt():
+    password = b"password"
+    salt = b"salt1234567890ab"
+    key = bcrypt(password, cost=10, salt=salt) # detected but not mapped
```
n1ckl0sk0rtge: **SHOULD FIX — orphan fixture; please delete the file.**

`grep -rn "Bcrypt\|bcrypt" python/src/main mapper/src/main engine/src/main` returns zero hits: the bcrypt rule was removed (M4) but this fixture stayed. It is the only `pycrypto` fixture with no matching `*Test.java`, so nothing loads it and no rule matches it.

The trailing comment is also inaccurate — `# detected but not mapped` implies detection happens; it does not.

README.md
```diff
@@ -36,13 +36,15 @@ It is part of **the [CBOMKit](https://github.com/cbomkit) toolset**.
 | Java     | [JCA](https://docs.oracle.com/javase/8/docs/technotes/guides/security/crypto/CryptoSpec.html) | 100%             |
 |          | [BouncyCastle](https://github.com/bcgit/bc-java) (*light-weight API*)                         | 100%[^1]         |
 | Python   | [pyca/cryptography](https://cryptography.io/en/latest/)                                       | 100%             |
+|          | [PyCryptodome(x)](https://www.pycryptodome.org/)                                              | 100%[^4]         |
```
n1ckl0sk0rtge: **SHOULD FIX — the 100% claim is not defensible against the shipped rule set.**

There is no rule for `Crypto.Protocol.HPKE` (the PR body's table still lists HPKE, and the only trace left is a dead translator case) and none for `Crypto.Protocol.SecretSharing`. Footnote 4's "also covers the legacy PyCrypto library" is also overstated — legacy `Crypto.Signature.PKCS1_v1_5`, `Crypto.Signature.PKCS1_PSS` and `Crypto.Cipher.XOR` module paths are absent; only the modern `pkcs1_15` / `pss` spellings are matched.

The PR body itself states tests cover "approximately 80% of the new detection rules".

```suggestion
|          | [PyCryptodome(x)](https://www.pycryptodome.org/)                                              | Partial[^4]      |
```

docs/DETECTION_RULE_STRUCTURE.md
```diff
@@ -94,6 +97,8 @@ In the tree of detected values, the values detected by these dependent detection
 
 At this point, you should have repeated all the steps starting from the `withMethodParameter` to here as many times as there are parameters in the function that you want to capture.
 
+The `withMethodParameter` section may be followed by one `withOtherParameters` indicating that the rule still matches even if an arbirary number of additional parameters follows. This feature should be used with care since it may lead to overlapping rules that cause duplicate detections. In languges like Python it allows the matching of combinations of optional parameters without analysing them in detail.
```
n1ckl0sk0rtge: **SUGGESTION — two typos in the new paragraph: "arbirary" and "languges".**

```suggestion
The `withMethodParameter` section may be followed by one `withOtherParameters` indicating that the rule still matches even if an arbitrary number of additional parameters follows. This feature should be used with care since it may lead to overlapping rules that cause duplicate detections. In languages like Python it allows the matching of combinations of optional parameters without analysing them in detail.
```

PR #506, opened 2026-08-18
feat(csharp): add C# ChaCha20Poly1305 detection rules 

Kaldesyvon:
Hello team,

After restoring C# I am adding new detection rules for ChaCha20Poly1305 algorithm.

This is my first PR in this project so I would appreciate any comments or help.

Thank you for reviewing.

Best,
Martin

### DIFF
--- a/csharp/src/main/java/com/ibm/plugin/rules/detection/dotnet/DotNetChaCha20Poly1305.java
+++ b/csharp/src/main/java/com/ibm/plugin/rules/detection/dotnet/DotNetChaCha20Poly1305.java
@@ -21,7 +21,9 @@ package com.ibm.plugin.rules.detection.dotnet;
 
 import com.ibm.engine.detection.MethodMatcher;
 import com.ibm.engine.language.csharp.tree.CSharpTree;
+import com.ibm.engine.model.CipherAction;
 import com.ibm.engine.model.context.CipherContext;
+import com.ibm.engine.model.factory.CipherActionFactory;
 import com.ibm.engine.model.factory.ValueActionFactory;
 import com.ibm.engine.rule.IDetectionRule;
 import com.ibm.engine.rule.builder.DetectionRuleBuilder;
@@ -39,7 +41,7 @@ public final class DotNetChaCha20Poly1305 {
                     .createDetectionRule()
                     .forObjectTypes(MethodMatcher.ANY)
                     .forMethods("Encrypt")
-                    .shouldBeDetectedAs(new ValueActionFactory<>("ENCRYPT"))
+                    .shouldBeDetectedAs(new CipherActionFactory<>(CipherAction.Action.ENCRYPT))
                     .withAnyParameters() // Byte[] or ReadOnlySpan<Byte> overloads
                     .buildForContext(new CipherContext())
                     .inBundle(() -> "DotNet")
@@ -51,7 +53,7 @@ public final class DotNetChaCha20Poly1305 {
                     .createDetectionRule()
                     .forObjectTypes(MethodMatcher.ANY)
                     .forMethods("Decrypt")
-                    .shouldBeDetectedAs(new ValueActionFactory<>("DECRYPT"))
+                    .shouldBeDetectedAs(new CipherActionFactory<>(CipherAction.Action.DECRYPT))
                     .withAnyParameters() // Byte[] or ReadOnlySpan<Byte> overloads
                     .buildForContext(new CipherContext())
                     .inBundle(() -> "DotNet")
--- a/csharp/src/main/java/com/ibm/plugin/rules/detection/dotnet/DotNetChaCha20Poly1305.java
+++ b/csharp/src/main/java/com/ibm/plugin/rules/detection/dotnet/DotNetChaCha20Poly1305.java
@@ -1,6 +1,6 @@
 /*
  * Sonar Cryptography Plugin
- * Copyright (C) 2024 PQCA
+ * Copyright (C) 2026 PQCA
  *
  * Licensed to the Apache Software Foundation (ASF) under one or more
  * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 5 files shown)


csharp/src/main/java/com/ibm/plugin/rules/detection/dotnet/DotNetChaCha20Poly1305.java
```diff
@@ -0,0 +1,50 @@
...
+        // nothing
+    }
+
+    private static final IDetectionRule<CSharpTree> CHACHA20POLY1305 =
+            new DetectionRuleBuilder<CSharpTree>()
+                    .createDetectionRule()
+                    .forObjectTypes("ChaCha20Poly1305")
+                    .forMethods("<init>")
+                    .shouldBeDetectedAs(new ValueActionFactory<>("CHACHA20POLY1305"))
+                    .withMethodParameter("[]byte")
```
fynnth: Hi @Kaldesyvon 
 "[]byte" isn't actually type-checked by the antlr parser right now, so this effectively only enforces "exactly 1 argument". Use .withAnyParameters() instead, with the comment // byte[] or ReadOnlySpan<Byte>, 1 param. 
I think like this it should work but not because it actually checks the type.

csharp/src/main/java/com/ibm/plugin/rules/detection/dotnet/DotNetChaCha20Poly1305.java
```diff
@@ -0,0 +1,50 @@
...
+                    .buildForContext(new CipherContext())
+                    .inBundle(() -> "DotNet")
+                    .withDependingDetectionRules(List.of());
```
fynnth: This is missing depending rules for the Encrypt()/Decrypt() calls ( the test file already calls them, but nothing detects it). The AES branch in our private repo has a good template for this with AesGcm/AesCcm:
 i thought along the lines of something similar:

private static final IDetectionRule<CSharpTree> CHACHA_ENCRYPT_OP =
        new DetectionRuleBuilder<CSharpTree>()
                .createDetectionRule()
                .forObjectTypes(MethodMatcher.ANY)
                .forMethods("Encrypt")
                .shouldBeDetectedAs(new ValueActionFactory<>("ENCRYPT"))
                .withAnyParameters()
                .buildForContext(new CipherContext())
                .inBundle(() -> "DotNet")
(plus an equivalent for Decrypt), then wire it in here instead of List.of(). That would give us findings like "ChaCha20-Poly1305 + Encrypt" or "ChaCha20-Poly1305 + Decrypt". and not only the bare constructor.

csharp/src/main/java/com/ibm/plugin/translation/translator/contexts/CSharpCipherContextTranslator.java
```diff
@@ -63,6 +59,8 @@ public final class CSharpCipherContextTranslator
             Optional<INode> result =
                     switch (valueStr) {
                         case "AES" -> Optional.of(new AES(detectionLocation));
+                        case "CHACHA20POLY1305" ->
```
fynnth, line 58: Once Encrypt/Decrypt detection is added above, we'll need matching cases here or those findings won't translate into anything:

case "ENCRYPT" -> Optional.of(new Encrypt(detectionLocation));
case "DECRYPT" -> Optional.of(new Decrypt(detectionLocation));

csharp/src/test/files/rules/detection/dotnet/DotNetChaCha20Poly1305TestFile.cs
```diff
@@ -0,0 +1,66 @@
+using System.Security.Cryptography;
+
+public class DotNetChaCha20Poly1305Test
+{
+    public void TestChaChaCreate()
+    {
+        var key = new byte [32];;
```
fynnth: double ;;

csharp/src/test/java/com/ibm/plugin/rules/detection/dotnet/DotNetChaCha20Poly1305Test.java
```diff
@@ -0,0 +1,64 @@
...
+import org.junit.jupiter.api.Test;
+
+class DotNetChaCha20Poly1305Test extends TestBase {
+    @Test
+    void test() throws Exception {
+        CSharpVerifier.verify("rules/detection/dotnet/DotNetChaCha20Poly1305TestFile.cs", this);
+    }
+
+    @Override
+    public void asserts(
```
fynnth, line 111: Once Encrypt/Decrypt findings are added, asserts() will need to differentiate by findingId (see DotNetAESComprehensiveTest in our private repo for the pattern, which maps findingId to expected combinations like "AES + Encrypt"). Right now every finding is checked against the same single assertion, which won't hold once the new Encrypt/Decrypt nodes show up.

csharp/src/main/java/com/ibm/plugin/rules/detection/dotnet/DotNetChaCha20Poly1305.java
```diff
@@ -32,16 +33,44 @@ private DotNetChaCha20Poly1305() {
         // nothing
     }
 
+    // chaCha20Poly1305.Encrypt(nonce, plaintext, ciphertext, tag [, associatedData])
+    private static final IDetectionRule<CSharpTree> CHACHA20POLY1305_ENCRYPT_OP =
+            new DetectionRuleBuilder<CSharpTree>()
+                    .createDetectionRule()
+                    .forObjectTypes(MethodMatcher.ANY)
+                    .forMethods("Encrypt")
+                    .shouldBeDetectedAs(new ValueActionFactory<>("ENCRYPT"))
```
san-zrl: If you detect this as

```
.shouldBeDetectedAs(new CipherActionFactory<>(CipherAction.Action.ENCRYPT))
```

you would not have to mix this into the handling of the value actions in the translator.

csharp/src/main/java/com/ibm/plugin/translation/translator/contexts/CSharpCipherContextTranslator.java
san-zrl: See my comment above. Compare with another translator handling encrypt/decrypt, e.g. python/src/main/java/com/ibm/plugin/translation/translator/contexts/PycaCipherContextTranslator.java

csharp/src/main/java/com/ibm/plugin/rules/detection/dotnet/DotNetChaCha20Poly1305.java
fynnth: Thats a good change thanks, will also implement that in my local rules :)

csharp/src/main/java/com/ibm/plugin/rules/detection/dotnet/DotNetChaCha20Poly1305.java
```diff
@@ -32,16 +33,44 @@ private DotNetChaCha20Poly1305() {
...
+                    .withAnyParameters() // Byte[] or ReadOnlySpan<Byte> overloads
```
san-zrl, line 45: This depends on how specific you want to be. withAnyParameters() means what is says - anything is ok. If I read the C# docs correctly this Encrypt requires 5 parameters. Your test now uses 4 which is OK if you have ANY here, but is this correct C#?

fynnth: This is correct c# code because AAD in this function has a default value that will be taken if only 4 parameters are there. That is also the reason for that specific test case and the test before that covers the 5 parameter option. 

HOWEVER i have tried to compile the file and it seems that `new ReadOnlySpan<byte>[12]; `creates an array of `ReadOnlySpan<byte>` which then fails once it gets used as parameter where it becomes non-convertible.
Also the file needs `using System;` for compilation, which doesnt matter for the test but just to not let it out. 

For me this works better: `Span<byte> nonce = stackalloc byte[12];`
Did it compile for you Martin?

san-zrl (dismissed):
Hi @Kaldesyvon - Thank you very much for the detection rule and the comprehensive tests. This looks good to me. @fynnth, what do you think?

fynnth (approved):

PR #509, opened 2026-08-19
Feat: withNamedMethodParameter() for python detection rules and tests

san-zrl:
# Add `withNamedMethodParameter` to the detection rule DSL

**Engine · Python · keyword-argument-aware parameter matching**

## Motivation

Python callers routinely pass arguments out of order using keyword syntax, and routinely omit optional parameters entirely:

```python
SHA512.new(truncate="256")                                       # data omitted
SHA512.new(truncate="256", data=b"msg")                          # reordered
PBKDF2HMAC(iterations=480000, algorithm=SHA256(), salt=s, length=32)  # fully reordered
```

The existing `withMethodParameter(type)` method is purely positional: it reads argument *n* for parameter *n* and requires an exact argument count enforced by `MethodMatcher`. It cannot handle reordered arguments, omitted optional parameters, or the keyword-argument call style that is idiomatic in Python cryptography APIs. `withNamedMethodParameter` solves all three.

## New DSL method

```java
ParametersFactoryBuilder<T> withNamedMethodParameter(
    @Nonnull String  name,      // Python keyword-argument name
    @Nonnull String  type,      // accepted type; use ANY for no constraint
             boolean optional   // true → absent or type-mismatched argument is silently skipped
);
```

### Ordering constraint

`withMethodParameter` declarations must precede all `withNamedMethodParameter` declarations in a rule. Violation throws `IllegalStateException` at rule construction time.

### MethodMatcher mode selection

When all parameters are declared with `withMethodParameter` the existing exact-count type-list `MethodMatcher` is used — no change in behaviour. As soon as at least one `withNamedMethodParameter` is declared the no-type-list constructor is used instead, which accepts any call to the matched method and delegates all structural checking to the extraction loop described below.

## Extraction algorithm

For each `withNamedMethodParameter` declaration, in declaration order:

1. **Keyword-name lookup** — scan the call-site argument list for a `RegularArgument` whose `keywordArgument().name()` equals `name`. If found, use that argument regardless of its position.
2. **Positional fallback** — if step 1 finds nothing, check the argument at the parameter's declaration index. Accept it only if that argument is itself positional (no keyword marker). If a preceding `optional=true` parameter was skipped, emit a `WARN` log before accepting (positional attribution may be ambiguous).
3. **Type check** — once an argument is resolved, verify its static type against the declared `type`. Skipped when `type` is `ANY`. When the type cannot be statically determined, `resolveTreeType` returns an accept-all predicate — the same conservative fallback used by `MethodMatcher` for positional parameters. On mismatch: `optional=true` → treat as absent and skip; `optional=false` → rule does not fire.
4. **Not found** — `optional=false`: rule does not fire. `optional=true`: skip silently.

## Pre-extraction guards (all run before any finding is emitted)

The following checks apply only when at least one named parameter is declared. All run before `MethodDetection` is emitted, so a rejected call produces no finding at all.

- **Unpacking guard** — if any call-site argument is an `UnpackingExpression` (`**d` or `*args`), the call is rejected. The contents of unpacked iterables and dicts cannot be statically inspected, so no structural guarantee can be made.
- **Unknown-keyword guard** — if any call-site keyword argument has a name that is not declared in any `withNamedMethodParameter` in the rule, the call is rejected. The call does not match the function signature the rule was written for.

## Example — `SHA512.new(data=None, truncate=None)`

```java
new DetectionRuleBuilder<Tree>()
    .createDetectionRule()
    .forObjectTypes("Crypto.Hash.SHA512", "Cryptodome.Hash.SHA512")
    .forMethods("new")
    .shouldBeDetectedAs(new ValueActionFactory<>("SHA512"))
    .withNamedMethodParameter("data",     ANY,   /*optional*/ true)
    .withNamedMethodParameter("truncate", "str", /*optional*/ true)
        .shouldBeDetectedAs(new AlgorithmParameterFactory<>(AlgorithmParameter.Kind.ANY))
        .asChildOfParameterWithId(0)
    .buildForContext(new DigestContext())
    .inBundle(() -> "PyCrypto")
```

| Call site | Matches? | `truncate` captured? | Reason |
|-----------|:--------:|:--------------------:|--------|
| `SHA512.new()` | ✓ | — | Both absent, both optional |
| `SHA512.new(data=b"msg")` | ✓ | — | data by keyword; truncate absent → skipped |
| `SHA512.new(truncate="256")` | ✓ | ✓ | truncate by keyword; data absent → skipped |
| `SHA512.new(data=b"msg", truncate="256")` | ✓ | ✓ | Both by keyword, canonical order |
| `SHA512.new(truncate="256", data=b"msg")` | ✓ | ✓ | Both by keyword, reordered |
| `SHA512.new(b"msg", "256")` | ✓ | ✓ | Both by positional fallback |
| `SHA512.new(b"msg", truncate="256")` | ✓ | ✓ | data positional fallback; truncate by keyword |
| `SHA512.new(truncate=256)` | ✓ | — | int literal fails "str" type check; optional → skipped |
| `SHA512.new(unknown_kwarg="x")` | ✗ | — | Unknown keyword guard |
| `SHA512.new(**d)` | ✗ | — | Unpacking guard |
| `SHA512.new(*args)` | ✗ | — | Unpacking guard |

## Changed files

| File | Change |
|------|--------|
| `engine/…/rule/Parameter.java` | Added `keywordName` and `keywordOptional` fields, two new constructors, and `getKeywordName()` / `isKeywordOptional()` accessors. |
| `engine/…/rule/DetectableParameter.java` | New constructor threading both new fields through `super()`. |
| `engine/…/rule/IDetectionRule.java` | Added `withNamedMethodParameter(name, type, optional)` to all six builder step interfaces. |
| `engine/…/rule/builder/DetectionRuleBuilderImpl.java` | Implemented the new method; enforces ordering constraint; updates `build()` to select the no-type-list `MethodMatcher` when named params are present. Introduced a `copy()` helper to eliminate repeated constructor calls. |
| `engine/…/language/python/PythonDetectionEngine.java` | Updated `analyseExpression()`: unpacking guard; unknown-keyword guard; `MethodDetection` emission moved after both guards; keyword-name lookup with positional fallback and Hole 1 `WARN` log; per-argument type check using `PythonSemantic.resolveTreeType`. Added `findArgumentByKeyword()` helper and extracted shared loop body into `processParameterExpression()`. Updated `getTraceSymbol()` to apply the same keyword-name lookup. |
| `python/…/detection/pycrypto/hash/PycryptoCryptoHash.java` | New example rule for `Crypto.Hash.SHA512.new` / `Cryptodome.Hash.SHA512.new` demonstrating the feature. |
| `python/…/detection/pycrypto/hash/PycryptoCryptoHashSHA512Test.java` + 15 Python fixture files | 15 test cases covering all positive call-site variants, type-mismatch skipping, unknown-keyword rejection, unpacking rejection, wrong-algorithm and wrong-method negative cases. |

## Known limitation

There is no upper-bound arity check: extra positional arguments beyond the declared parameter count are silently ignored. This does not occur in valid Python (the interpreter would raise `TypeError`), so it is not a source of false positives in practice. The unpacking guard already handles the one realistic case where extra arguments could arrive opaquely (`**d`).

### DIFF
--- a/engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
+++ b/engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
@@ -32,14 +32,20 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.sonar.plugins.python.api.PythonCheck;
 import org.sonar.plugins.python.api.PythonVisitorContext;
 import org.sonar.plugins.python.api.symbols.Symbol;
 import org.sonar.plugins.python.api.tree.*;
 
 public class PythonDetectionEngine implements IDetectionEngine<Tree, Symbol> {
+    private static final Logger LOG = LoggerFactory.getLogger(PythonDetectionEngine.class);
+
     @Nonnull
     private final DetectionStore<PythonCheck, Tree, Symbol, PythonVisitorContext> detectionStore;
 
@@ -351,6 +357,29 @@ public class PythonDetectionEngine implements IDetectionEngine<Tree, Symbol> {
     @Nonnull
     private Optional<TraceSymbol<Symbol>> getTraceSymbol(
             @Nonnull Parameter<Tree> parameter, @Nonnull List<Argument> arguments) {
+        Optional<String> keywordName = parameter.getKeywordName();
+        if (keywordName.isPresent()) {
+            // Named parameter: keyword-name lookup first, then positional fallback
+            Optional<Argument> resolved =
+                    findArgumentByKeyword(keywordName.get(), parameter.getIndex(), arguments);
+            if (resolved.isEmpty()) {
+                if (parameter.isKeywordOptional()) {
+                    return Optional.of(TraceSymbol.createWithStateDifferent());
+                }
+                return Optional.of(TraceSymbol.createWithStateDifferent());
+            }
+            Argument arg = resolved.get();
+            if (arg instanceof RegularArgument regularArg) {
+                Expression expressionArg = regularArg.expression();
+                if (expressionArg.is(Tree.Kind.NAME)) {
+                    Name nameArg = (Name) expressionArg;
+                    return Optional.of(TraceSymbol.createFrom(nameArg.symbol()));
+                }
+            }
+            return Optional.of(TraceSymbol.createWithStateNoSymbol());
+        }
+
+        // Positional parameter (original behaviour)
         if (parameter.getIndex() >= arguments.size()) {
             return Optional.of(TraceSymbol.createWithStateDifferent());
         }
@@ -414,10 +443,6 @@ public class PythonDetectionEngine implements IDetectionEngine<Tree, Symbol> {
         }
 
         DetectionRule<Tree> detectionRule = (DetectionRule<Tree>) detectionStore.getDetectionRule();
-        if (detectionRule.actionFactory() != null) {
-            MethodDetection<Tree> methodDetection = new MethodDetection<>(expressionTree, null);
-            detectionStore.onReceivingNewDetection(methodDetection);
-        }
 
         // Extracts the arguments for the provided expression
         List<Argument> arguments = expressionTree.arguments();
@@ -427,57 +452,211 @@ public class PythonDetectionEngine implements IDetectionEngine<Tree, Symbol> {
         // TODO: It would be better to have a case disjunction to use either
         // isInvocationOnVariable or isInitForVariable, but it is difficult in Python
 
-        int index = 0;
+        // When named parameters are present, reject any call site that passes a keyword argument
+        // whose name is not declared in the rule. Such a call cannot match the function signature
+        // the rule was written for, so it must not fire. This check must happen before emitting
+        // MethodDetection so that rejected calls produce no finding at all.
+        // Positional-only rules are unaffected: the MethodMatcher's exact-count check already
+        // rejects wrong-signature calls before we reach this point.
+        boolean hasNamedParams =
+                detectionRule.parameters().stream().anyMatch(p -> p.getKeywordName().isPresent());
+        if (hasNamedParams) {
+            // Reject if any argument is a dict-unpacking expression (**d) or a sequence-unpacking
+            // expression (*args). We cannot statically inspect what keys or values they contain,
+            // so we cannot verify the call matches the rule's declared parameters.
+            boolean hasUnpackingArg =
+                    arguments.stream().anyMatch(a -> !(a instanceof RegularArgument));
+            if (hasUnpackingArg) {
+                return;
+            }
+
+            // Reject if any keyword argument name is not declared in the rule.
+            Set<String> declaredKeywordNames =
+                    detectionRule.parameters().stream()
+                            .flatMap(p -> p.getKeywordName().stream())
+                            .collect(Collectors.toSet());
+            boolean hasUnknownKeyword =
+                    arguments.stream()
+                            .anyMatch(
+                                    a ->
+                                            a instanceof RegularArgument ra
+                                                    && ra.keywordArgument() != null
+                                                    && !declaredKeywordNames.contains(
+                                                            ra.keywordArgument().name()));
+            if (hasUnknownKeyword) {
+                return;
+            }
+        }
+
+        if (detectionRule.actionFactory() != null) {
+            MethodDetection<Tree> methodDetection = new MethodDetection<>(expressionTree, null);
+            detectionStore.onReceivingNewDetection(methodDetection);
+        }
+
+        // Track whether a preceding optional named parameter was skipped (Hole 1 mitigation).
+        boolean precedingOptionalNamedParamWasSkipped = false;
+
+        int positionalIndex = 0; // tracks the index among all parameters (for positional fallback)
         for (Parameter<Tree> parameter : detectionRule.parameters()) {
-            if (!checkCurrentIndexState(
-                    index, arguments, isInvocation, traceSymbol, expressionTree)) {
-                index++;
-                continue;
-            }
-            // the expression tree of the parameter
-            Tree expression = arguments.get(index); // this is an Argument tree
-            if (expression instanceof RegularArgument regularArgument) {
-                expression = regularArgument.expression();
-            }
+            Optional<String> keywordName = parameter.getKeywordName();
 
-            /*
-             * This method resolves the detection parameter in an inner scope.
-             * If unsuccessful, it falls back to resolving values in the outer scope using the provided expression and detectableParameter.
-             */
-            if (parameter.is(DetectableParameter.class)) {
-                DetectableParameter<Tree> detectableParameter =
-                        (DetectableParameter<Tree>) parameter;
-                // try to resolve value in inner scope
-                List<ResolvedValue<Object, Tree>> resolvedValues =
-                        resolveValuesInInnerScope(
-                                Object.class, expression, detectableParameter.getiValueFactory());
-                if (resolvedValues.isEmpty()) {
-                    // goto outer scope
-                    resolveValuesInOuterScope(expression, detectableParameter);
-                } else {
-                    resolvedValues.stream()
-                            .map(
-                                    resolvedValue ->
-                                            new ValueDetection<>(
-                                                    resolvedValue,
-                                                    detectableParameter,
-                                                    expressionTree,
-                                                    expressionTree))
-                            .forEach(detectionStore::onReceivingNewDetection);
+            if (keywordName.isPresent()) {
+                // --- Named parameter extraction (proposal §3) ---
+                Optional<Argument> resolvedArg =
+                        findArgumentByKeyword(keywordName.get(), positionalIndex, arguments);
+
+                if (resolvedArg.isEmpty()) {
+                    if (parameter.isKeywordOptional()) {
+                        // optional parameter absent → skip silently
+                        precedingOptionalNamedParamWasSkipped = true;
+                        positionalIndex++;
+                        continue;
+                    } else {
+                        // required named parameter absent → rule does not fire; stop processing
+                        return;
+                    }
                 }
-            } else if (!parameter.getDetectionRules().isEmpty()) {
-                /*
-                 * This case is reached when the parameter is not a DetectableParameter (the rule does not contains `.shouldBeDetectedAs`),
-                 * but has depending detection rules (the rule contain `.addDependingDetectionRules`).
-                 * This happens usually for parameters that are intermediary function, that we have to resolve but we don't want to capture their value.
-                 * In this case, we resolve the parameter with the depending detection rule with an EXPRESSION scope,
-                 * this way we ensure to only resolve the right parameter content and not similar calls in the same function scope.
-                 */
-                detectionStore.onDetectedDependingParameter(
-                        parameter, expression, DetectionStore.Scope.EXPRESSION);
+
+                // Check whether we used positional fallback AND a preceding optional was skipped
+                // (Hole 1 warning — proposal §5).
+                Argument arg = resolvedArg.get();
+                boolean usedPositionalFallback =
+                        (arg instanceof RegularArgument ra && ra.keywordArgument() == null);
+                if (usedPositionalFallback && precedingOptionalNamedParamWasSkipped) {
+                    LOG.warn(
+                            "Named parameter \"{}\" resolved via positional fallback at index {},"
+                                    + " but a preceding optional named parameter was not found."
+                                    + " The positional argument may belong to a different parameter."
+                                    + " Rule: {}, location: {}",
+                            keywordName.get(),
+                            positionalIndex,
+                            detectionStore.getDetectionRule().getClass().getSimpleName(),
+                            expressionTree.firstToken().value());
+                }
+
+                // extract expression from the resolved argument
+                Tree expression = arg;
+                if (expression instanceof RegularArgument regularArgument) {
+                    expression = regularArgument.expression();
+                }
+
+                // Type check: verify the resolved argument's type against the declared parameter
+                // type. Skipped when the declared type is ANY (wildcard). When the type cannot be
+                // statically resolved, resolveTreeType returns an accept-all IType — the same
+                // conservative behaviour used by MethodMatcher for positional parameters.
+                if (!parameter.getParameterType().equals(MethodMatcher.ANY)) {
+                    Optional<IType> resolvedType = PythonSemantic.resolveTreeType(expression);
+                    boolean typeMatches =
+                            resolvedType.map(t -> t.is(parameter.getParameterType())).orElse(true);
+                    if (!typeMatches) {
+                        if (parameter.isKeywordOptional()) {
+                            // Type mismatch on an optional param: treat as absent and skip.
+                            precedingOptionalNamedParamWasSkipped = true;
+                            positionalIndex++;
+                            continue;
+                        } else {
+                            // Type mismatch on a required param: rule does not fire.
+                            return;
+                        }
+                    }
+                }
+
+                if (!checkCurrentIndexState(
+                        positionalIndex, arguments, isInvocation, traceSymbol, expressionTree)) {
+                    positionalIndex++;
+                    continue;
+                }
+
+                processParameterExpression(parameter, expression, expressionTree);
+            } else {
+                // --- Positional parameter extraction (original behaviour) ---
+                if (!checkCurrentIndexState(
+                        positionalIndex, arguments, isInvocation, traceSymbol, expressionTree)) {
+                    positionalIndex++;
+                    continue;
+                }
+                Tree expression = arguments.get(positionalIndex);
+                if (expression instanceof RegularArgument regularArgument) {
+                    expression = regularArgument.expression();
+                }
+
+                processParameterExpression(parameter, expression, expressionTree);
             }
 
-            index++;
+            positionalIndex++;
+        }
+    }
+
+    /**
+     * Tries to find an argument matching the given keyword name. Falls back to the positional index
+     * if no keyword-named argument is found and the argument at that index is positional (i.e. has
+     * no keyword name itself).
+     *
+     * @param keywordName the name to look for as a keyword argument
+     * @param positionalIndex the fallback positional index
+     * @param arguments the full argument list at the call site
+     * @return the matching argument, or empty if neither keyword nor positional match is available
+     */
+    @Nonnull
+    private Optional<Argument> findArgumentByKeyword(
+            @Nonnull String keywordName, int positionalIndex, @Nonnull List<Argument> arguments) {
+        // Step 1: keyword-name lookup
+        for (Argument arg : arguments) {
+            if (arg instanceof RegularArgument ra
+                    && ra.keywordArgument() != null
+                    && keywordName.equals(ra.keywordArgument().name())) {
+                return Optional.of(arg);
+            }
+        }
+        // Step 2: positional fallback — accept only if the argument at positionalIndex is itself
+        // positional (no keyword marker) to avoid misattributing a keyword arg to the wrong param.
+        if (positionalIndex < arguments.size()) {
+            Argument arg = arguments.get(positionalIndex);
+            if (arg instanceof RegularArgument ra && ra.keywordArgument() == null) {
+                return Optional.of(arg);
+            }
+        }
+        return Optional.empty();
+    }
+
+    /**
+     * Shared logic for processing a resolved parameter expression (both positional and named
+     * parameters). Mirrors the inner body of the original analyseExpression loop.
+     */
+    private void processParameterExpression(
+            @Nonnull Parameter<Tree> parameter,
+            @Nonnull Tree expression,
+            @Nonnull CallExpression expressionTree) {
+        if (parameter.is(DetectableParameter.class)) {
+            DetectableParameter<Tree> detectableParameter = (DetectableParameter<Tree>) parameter;
+            // try to resolve value in inner scope
+            List<ResolvedValue<Object, Tree>> resolvedValues =
+                    resolveValuesInInnerScope(
+                            Object.class, expression, detectableParameter.getiValueFactory());
+            if (resolvedValues.isEmpty()) {
+                // goto outer scope
+                resolveValuesInOuterScope(expression, detectableParameter);
+            } else {
+                resolvedValues.stream()
+                        .map(
+                                resolvedValue ->
+                                        new ValueDetection<>(
+                                                resolvedValue,
+                                                detectableParameter,
+                                                expressionTree,
+                                                expressionTree))
+                        .forEach(detectionStore::onReceivingNewDetection);
+            }
+        } else if (!parameter.getDetectionRules().isEmpty()) {
+            /*
+             * This case is reached when the parameter is not a DetectableParameter (the rule does not contains `.shouldBeDetectedAs`),
+             * but has depending detection rules (the rule contain `.addDependingDetectionRules`).
+             * This happens usually for parameters that are intermediary function, that we have to resolve but we don't want to capture their value.
+             * In this case, we resolve the parameter with the depending detection rule with an EXPRESSION scope,
+             * this way we ensure to only resolve the right parameter content and not similar calls in the same function scope.
+             */
+            detectionStore.onDetectedDependingParameter(
+                    parameter, expression, DetectionStore.Scope.EXPRESSION);
         }
     }
 
--- a/engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
+++ b/engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
@@ -363,9 +363,6 @@ public class PythonDetectionEngine implements IDetectionEngine<Tree, Symbol> {
             Optional<Argument> resolved =
                     findArgumentByKeyword(keywordName.get(), parameter.getIndex(), arguments);
             if (resolved.isEmpty()) {
-                if (parameter.isKeywordOptional()) {
-                    return Optional.of(TraceSymbol.createWithStateDifferent());
-                }
                 return Optional.of(TraceSymbol.createWithStateDifferent());
             }
             Argument arg = resolved.get();
@@ -486,6 +483,22 @@ public class PythonDetectionEngine implements IDetectionEngine<Tree, Symbol> {
             if (hasUnknownKeyword) {
                 return;
             }
+
+            // Reject if any required named parameter is absent from the call site.
+            // This must run before MethodDetection is emitted so that a call that is missing a
+            // required argument produces no finding at all (not just a missing child detection).
+            int preCheckIndex = 0;
+            for (Parameter<Tree> p : detectionRule.parameters()) {
+                if (p.getKeywordName().isPresent() && !p.isKeywordOptional()) {
+                    Optional<Argument> present =
+                            findArgumentByKeyword(
+                                    p.getKeywordName().get(), preCheckIndex, arguments);
+                    if (present.isEmpty()) {
+                        return;
+                    }
+                }
+                preCheckIndex++;
+            }
         }
 
         if (detectionRule.actionFactory() != null) {
@@ -512,7 +525,9 @@ public class PythonDetectionEngine implements IDetectionEngine<Tree, Symbol> {
                         positionalIndex++;
                         continue;
                     } else {
-                        // required named parameter absent → rule does not fire; stop processing
+                        // required named parameter absent → stop processing
+                        // (pre-flight at line 490 already guarantees this branch is unreachable
+                        // for named parameters; kept as a defensive fallback)
                         return;
                     }
                 }
@@ -561,8 +576,7 @@ public class PythonDetectionEngine implements IDetectionEngine<Tree, Symbol> {
                     }
                 }
 
-                if (!checkCurrentIndexState(
-                        positionalIndex, arguments, isInvocation, traceSymbol, expressionTree)) {
+                if (!checkSymbolTraceState(isInvocation, traceSymbol, expressionTree)) {
                     positionalIndex++;
                     continue;
                 }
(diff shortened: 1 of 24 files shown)


n1ckl0sk0rtge (commented):
Inline comments below. Summary and verdict in the follow-up review.

engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
```diff
@@ -427,57 +452,211 @@ private void analyseExpression(
...
+                            positionalIndex++;
+                            continue;
+                        } else {
+                            // Type mismatch on a required param: rule does not fire.
+                            return;
+                        }
+                    }
+                }
+
+                if (!checkCurrentIndexState(
```
n1ckl0sk0rtge: **Blocker — the main example of this PR does not work.**

`checkCurrentIndexState` still starts with `if (arguments.size() <= index) return false;`, and here we pass the *declared* index. For a named parameter the argument was already resolved, possibly at a lower position, so this check throws the result away.

Walk through `SHA512.new(truncate="256")` — one argument in the call:

1. `data` (index 0): no `data=` keyword; the argument at position 0 is `truncate="256"`, which has a keyword, so it is not taken as positional. `data` is optional, so it is skipped and `positionalIndex` becomes 1.
2. `truncate` (index 1): found by name at position 0. Good.
3. This line: `checkCurrentIndexState(1, ...)` -> `arguments.size() (1) <= 1` -> returns `false` -> the parameter is dropped.

I verified it by printing the child stores and running each positive test alone:

| call site | `truncate` captured? |
|---|---|
| `SHA512.new(truncate="256")` | **no (0 children)** |
| `SHA512.new(data=b"msg", truncate="256")` | yes |
| `SHA512.new(truncate="256", data=b"msg")` | yes |
| `SHA512.new(b"msg", "256")` | yes |
| `SHA512.new(b"msg", truncate="256")` | yes |
| Cryptodome variant | yes |

This is exactly the case the PR is about: an optional parameter left out, a later one passed by name.

Suggested fix: for named parameters the argument is already resolved, so the arity part of the check is pointless. Either pass the real position of the resolved argument, or split `checkCurrentIndexState` so the `arguments.size() <= index` part is skipped for named parameters.

engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
```diff
@@ -427,57 +452,211 @@ private void analyseExpression(
...
-                            .map(
-                                    resolvedValue ->
-                                            new ValueDetection<>(
-                                                    resolvedValue,
-                                                    detectableParameter,
-                                                    expressionTree,
-                                                    expressionTree))
-                            .forEach(detectionStore::onReceivingNewDetection);
+        if (detectionRule.actionFactory() != null) {
+            MethodDetection<Tree> methodDetection = new MethodDetection<>(expressionTree, null);
```
n1ckl0sk0rtge, line 515: **Blocker — `optional=false` does not stop the rule from firing.**

The finding is created here, before the extraction loop. The required-parameter paths below only `return` afterwards, so the `MethodDetection` is already in the store.

I tested it: I changed the rule to `.withNamedMethodParameter("truncate", "str", /* optional */ false)` and ran `testNoArgs` (`SHA512.new()`, no `truncate` at all). The test still passes, which means the `# Noncompliant` finding is still reported.

So `optional=false` only removes the captured child value. It does not prevent detection, which is what the PR description promises.

Either fix the description, or move the required-parameter checks up into the guard block above, next to the unpacking and unknown-keyword guards.

engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
```diff
@@ -427,57 +452,211 @@ private void analyseExpression(
...
+                        findArgumentByKeyword(keywordName.get(), positionalIndex, arguments);
+
+                if (resolvedArg.isEmpty()) {
+                    if (parameter.isKeywordOptional()) {
+                        // optional parameter absent → skip silently
+                        precedingOptionalNamedParamWasSkipped = true;
+                        positionalIndex++;
+                        continue;
+                    } else {
+                        // required named parameter absent → rule does not fire; stop processing
```
n1ckl0sk0rtge: See the comment on the `MethodDetection` emission above. This `return` runs *after* the finding was already created, so a missing required parameter does not prevent the rule from firing — it only skips the child value.

engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
```diff
@@ -427,57 +452,211 @@ private void analyseExpression(
...
+            if (hasUnpackingArg) {
+                return;
             }
-            // the expression tree of the parameter
-            Tree expression = arguments.get(index); // this is an Argument tree
-            if (expression instanceof RegularArgument regularArgument) {
-                expression = regularArgument.expression();
+
+            // Reject if any keyword argument name is not declared in the rule.
+            Set<String> declaredKeywordNames =
```
n1ckl0sk0rtge: `declaredKeywordNames` only collects names from `withNamedMethodParameter`. Parameters declared with `withMethodParameter` have no name, so they are never in this set.

That means a mixed rule rejects legal Python. Example:

```java
.withMethodParameter("str")                       // no name
.withNamedMethodParameter("length", "int", true)
```

```python
f(algorithm="sha256", length=32)   # rejected: "algorithm" is an "unknown keyword"
```

Passing a positional parameter by keyword is completely normal Python, so every mixed rule has this hole. Worth deciding whether mixed rules are supported at all — if yes, positional declarations need a name too.

engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
```diff
@@ -351,6 +357,29 @@ public Optional<TraceSymbol<Symbol>> getNewClassParameterSymbol(
...
     private Optional<TraceSymbol<Symbol>> getTraceSymbol(
             @Nonnull Parameter<Tree> parameter, @Nonnull List<Argument> arguments) {
+        Optional<String> keywordName = parameter.getKeywordName();
+        if (keywordName.isPresent()) {
+            // Named parameter: keyword-name lookup first, then positional fallback
+            Optional<Argument> resolved =
+                    findArgumentByKeyword(keywordName.get(), parameter.getIndex(), arguments);
+            if (resolved.isEmpty()) {
+                if (parameter.isKeywordOptional()) {
+                    return Optional.of(TraceSymbol.createWithStateDifferent());
```
n1ckl0sk0rtge: Dead code: both branches return the same value.

```java
if (parameter.isKeywordOptional()) {
    return Optional.of(TraceSymbol.createWithStateDifferent());
}
return Optional.of(TraceSymbol.createWithStateDifferent());
```

Either the optional case was meant to return something else, or the `if` can go.

engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
```diff
@@ -427,57 +452,211 @@ private void analyseExpression(
...
+                    }
+                }
+
+                // Check whether we used positional fallback AND a preceding optional was skipped
+                // (Hole 1 warning — proposal §5).
+                Argument arg = resolvedArg.get();
+                boolean usedPositionalFallback =
+                        (arg instanceof RegularArgument ra && ra.keywordArgument() == null);
+                if (usedPositionalFallback && precedingOptionalNamedParamWasSkipped) {
+                    LOG.warn(
```
n1ckl0sk0rtge: Two problems with this log:

1. It runs once per matching call site during a real scan, at WARN level. That will be noisy in SonarQube analysis logs.
2. Two of the values are not what they look like:
   - `detectionStore.getDetectionRule().getClass().getSimpleName()` is always the string `"DetectionRule"`, so it does not identify the rule.
   - `expressionTree.firstToken().value()` is the token text (`"SHA512"`), not a location.

Suggestion: drop it, or make it `LOG.debug` and log a real line number.

engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
```diff
@@ -427,57 +452,211 @@ private void analyseExpression(
...
-                            .forEach(detectionStore::onReceivingNewDetection);
+        if (detectionRule.actionFactory() != null) {
+            MethodDetection<Tree> methodDetection = new MethodDetection<>(expressionTree, null);
+            detectionStore.onReceivingNewDetection(methodDetection);
+        }
+
+        // Track whether a preceding optional named parameter was skipped (Hole 1 mitigation).
+        boolean precedingOptionalNamedParamWasSkipped = false;
+
+        int positionalIndex = 0; // tracks the index among all parameters (for positional fallback)
```
n1ckl0sk0rtge: `positionalIndex` is always equal to `parameter.getIndex()` — the builder sets `index = parameters.size()` when each parameter is added, so they cannot drift apart. The extra counter can go, which also removes the risk of the two getting out of sync when someone edits the `continue` paths later.

(`getTraceSymbol` above already uses `parameter.getIndex()` for the same purpose.)

engine/src/main/java/com/ibm/engine/rule/builder/DetectionRuleBuilderImpl.java
```diff
@@ -485,11 +354,21 @@ private IDetectionRule<T> build() {
...
-                            this.methodNames,
-                            this.parameters.stream().map(Parameter::getParameterType).toList());
+            // If any parameter is a named parameter, use the no-type-list MethodMatcher so that
+            // the matcher accepts any call to the method and delegates structural matching to
+            // the extraction loop in the engine (proposal §2.1).
+            final boolean hasNamedParams =
+                    parameters.stream().anyMatch(p -> p.getKeywordName().isPresent());
+            final MethodMatcher<T> methodMatcher;
+            if (hasNamedParams) {
+                methodMatcher = new MethodMatcher<>(this.objectTypes, this.methodNames);
```
n1ckl0sk0rtge, line 364: One named parameter removes the type list for the **whole** rule, but the extraction loop in `PythonDetectionEngine` only type-checks named parameters. So positional parameters in a mixed rule lose both their type check and the exact-arity check.

```java
.withMethodParameter("str")                        // type checked nowhere now
.withNamedMethodParameter("length", "int", true)
```

Before: `MethodMatcher` checked `"str"` and required an exact argument count.
After: `f(12345, length=32)` matches, even though the first parameter should be a string.

Either type-check positional parameters in the loop as well, or keep a type list for the positional prefix.

engine/src/main/java/com/ibm/engine/rule/builder/DetectionRuleBuilderImpl.java
```diff
@@ -516,29 +395,58 @@ private void checkDetectionParameterState() {
                                 parameterShouldMatchExactTypes,
                                 detectionRules));
```
n1ckl0sk0rtge, line 396: The keyword fields are dropped in this branch. The other two branches below pass `pendingKeywordName` / `pendingKeywordOptional`, but this one (`buildingNewDetectionParameter && iValueFactory == null`) does not.

That branch is reached by the depending-rules path, which the DSL now exposes:

```java
.withNamedMethodParameter("algorithm", ANY, false)
.addDependingDetectionRules(PycaHash.rules())      // name silently thrown away
```

The rule then falls back to positional matching with no error and no warning.

engine/src/main/java/com/ibm/engine/rule/builder/DetectionRuleBuilderImpl.java
```diff
@@ -125,239 +139,121 @@ public IDetectionRule.NameBuilder<T> forObjectTypes(@Nonnull String... types) {
...
-                positionMove,
-                parameterShouldMatchExactTypes,
-                buildingNewDetectionParameter,
-                bundle);
+        return copy();
+    }
+
+    @Nonnull
+    @Override
+    public IDetectionRule.ParametersFactoryBuilder<T> withNamedMethodParameter(
```
n1ckl0sk0rtge, line 217: Two small things worth adding here:

- `name` is not validated. An empty string is accepted.
- Duplicate names are accepted. Two parameters declared with the same name would both resolve to the same call-site argument.

Both are cheap to catch at rule construction time, like the ordering constraint above.

engine/src/main/java/com/ibm/engine/rule/Parameter.java
```diff
@@ -29,17 +31,27 @@ public class Parameter<T> {
     protected boolean shouldMatchExactTypes;
     protected final int index;
 
+    /** Non-null when this parameter was declared with {@code withNamedMethodParameter}. */
+    @Nullable private final String keywordName;
```
n1ckl0sk0rtge, line 35: I would not put two loose fields on the base class. Here is what we could do instead.

|  | plain | named |
|---|---|---|
| not detectable | `Parameter` | `NamedParameter` |
| detectable | `DetectableParameter` | `NamedDetectableParameter` |

And this PR already creates both named variants: a plain named `Parameter` for the depending-rules path, and a named `DetectableParameter` for the value path.

There is a second problem: `Parameter.is()` compares the type tag with `equals`, not `instanceof`. A subclass of `DetectableParameter` would either fail `is(DetectableParameter.class)`, or have to pass the parent tag and lie about its real class. The engine would then need `instanceof` in some places and `is()` in others.

**Nonnull concern: ** This state is representable and means nothing:

```java
new Parameter<>(type, 0, false, rules, null, true);   // no name, but "optional" = true
```

Also the new public constructor below copies all five field assignments instead of delegating to the protected one.

**Suggestion — one nullable field holding a small record:**

```java
public record NamedArgument(@Nonnull String name, boolean optional) {}

@Nullable private final NamedArgument namedArgument;

@Nonnull
public Optional<NamedArgument> getNamedArgument() {
    return Optional.ofNullable(namedArgument);
}
```

- the name is non-null by construction, enforced in one place
- `null` means "not a named parameter" — the meaningless combination disappears
- one new constructor instead of two, and the existing ones can delegate
- the engine reads one field instead of two

I would also rename `keyword*` to `argument*`. Named arguments exist in several languages, and `Parameter` is the shared language-neutral class — `keyword` is Python wording.

engine/src/main/java/com/ibm/engine/rule/IDetectionRule.java
```diff
@@ -72,6 +72,10 @@ interface ActionDetectionBuilder<T> {
 
         @Nonnull
         ParametersFactoryBuilder<T> withMethodParameterMatchExactType(@Nonnull String type);
+
+        @Nonnull
+        ParametersFactoryBuilder<T> withNamedMethodParameter(
```
n1ckl0sk0rtge, line 77: This method sits on `IDetectionRule<T>`, so a Java or Go rule can call it too. But the feature is Python-only:

- `JavaDetectionEngine` and the Go engine never read `getKeywordName()` — I grepped, there are no callers there.
- `DetectionRuleBuilderImpl.build()` still drops the `MethodMatcher` type list for any rule with a named parameter, whatever the language.

Combined, a Java rule using this would match any call with that method name and any argument count, and then read arguments by position. That is a false-positive generator.

Suggestion: reject named parameters for non-Python rules, or at minimum add a clear warning in the javadoc here.

python/src/main/java/com/ibm/plugin/rules/detection/pycrypto/hash/PycryptoCryptoHash.java
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * <ul>
+ *   <li>{@code data} — optional bytes to hash; ANY type, optional
+ *   <li>{@code truncate} — optional truncation variant ("256" or "224"); type {@code str}, optional
+ * </ul>
+ *
+ * <p>Both parameters are declared with {@code withNamedMethodParameter} so that keyword-argument
+ * call sites (e.g. {@code SHA512.new(truncate="256")}) are handled correctly regardless of argument
+ * order. The {@code truncate} value is captured as an {@link AlgorithmParameter} so that the
+ * truncated variant (SHA-512/256 or SHA-512/224) can be reported.
```
n1ckl0sk0rtge: This claim is not delivered yet. Even in the tests where `truncate="256"` **is** captured, the translated node is:

```
SHA-512 -> [Oid, Digest, BlockSize, DigestSize]
```

There is no `SHA-512/256` and no `SHA-512/224`. The captured `AlgorithmParameter` never reaches the output model.

Either add the mapper support, or drop this sentence so the rule does not promise more than it does.

python/src/main/java/com/ibm/plugin/rules/detection/pycrypto/hash/PycryptoCryptoHash.java
```diff
@@ -0,0 +1,91 @@
...
+    private static final Supplier<List<IDetectionRule<Tree>>> RULES =
+            Memoize.of(PycryptoCryptoHash::buildRules);
+
+    @Nonnull
+    public static List<IDetectionRule<Tree>> rules() {
+        return RULES.get();
+    }
+
+    @Nonnull
+    private static List<IDetectionRule<Tree>> buildRules() {
```
n1ckl0sk0rtge, line 88: This rule is not registered anywhere. `PythonDetectionRules.buildRules()` does not list `PycryptoCryptoHash`, and I found no other reference outside this class's own test.

So in a real scan this rule never runs. Either add it to `PythonDetectionRules`, or move the example to test scope so it is clear it is a demo and not shipped detection.

python/src/test/java/com/ibm/plugin/rules/detection/pycrypto/hash/PycryptoCryptoHashSHA512Test.java
```diff
@@ -0,0 +1,243 @@
...
+        assertThat(digestNode.asString()).isEqualTo("DIGEST");
+
+        /*
+         * If a truncate child detection store is present, validate its value.
+         * Tests that pass truncate="256" produce one AlgorithmParameter child.
+         * Tests that do NOT pass truncate produce no AlgorithmParameter child.
+         */
+        DetectionStore<PythonCheck, Tree, Symbol, PythonVisitorContext> truncateStore =
+                getStoreOfValueType(AlgorithmParameter.class, detectionStore.getChildren());
+        if (truncateStore != null) {
```
n1ckl0sk0rtge: **This `if` is why the suite is green while the feature is broken.**

If nothing was captured, `truncateStore` is `null`, the block is skipped, and the test passes. See my comment on `PythonDetectionEngine` — `SHA512.new(truncate="256")` captures nothing today, and this assert does not notice.

The `# Noncompliant` comments do not help either, because every fixture uses the same expected message:

```python
h = SHA512.new(truncate="256")   # Noncompliant {{(MessageDigest) SHA-512}}
h = SHA512.new(data=b"msg")      # Noncompliant {{(MessageDigest) SHA-512}}
```

The message says nothing about `truncate`.

As it stands, 6 of the 10 positive tests only prove "SHA-512 was found", which the old positional code could already do.

Suggestion: let each test declare whether it expects the child, then assert it:

```java
assertThat(truncateStore).isNotNull();                     // capture expected
assertThat(truncateValue.asString()).isEqualTo("256");
```

```java
assertThat(truncateStore).isNull();                        // no capture expected
```

python/src/test/java/com/ibm/plugin/rules/detection/pycrypto/hash/PycryptoCryptoHashSHA512Test.java
```diff
@@ -0,0 +1,243 @@
...
+    void testDataKeyword() {
+        PythonCheckVerifier.verify(BASE_PATH + "SHA512NewDataKeywordTest.py", this);
+    }
+
+    /**
+     * SHA512.new(truncate="256") — truncate by keyword, data absent; truncate child must be
+     * captured.
+     */
+    @Test
+    void testTruncateKeywordOnly() {
```
n1ckl0sk0rtge, line 84: This is the test for the PR's headline example, and today it passes without capturing anything — see the comment on the conditional assert below.

Once that assert is made unconditional, this test will fail until the arity bug in `PythonDetectionEngine` is fixed. That is the right order: make the test fail first, then fix the engine.

n1ckl0sk0rtge (changes requested):
**Blockers**

1. `SHA512.new(truncate="256")` — the headline example — captures nothing. The arity check in `checkCurrentIndexState` still uses the declared index, so a keyword argument at a lower position is dropped.
2. The tests cannot catch that. `asserts()` wraps the check in `if (truncateStore != null)`, and all fixtures expect the same `# Noncompliant` message, so 6 of the 10 positive tests assert nothing about the new feature.
3. `optional=false` still reports a finding — `MethodDetection` is emitted before the required-parameter checks. Verified by making `truncate` required and running `SHA512.new()`.
4. Mixed positional + named rules lose the positional type and arity checks, and the unknown-keyword guard then rejects legal Python.
5. `withNamedMethodParameter` followed by `addDependingDetectionRules` silently loses the name.

**Worth deciding**

The DSL sits on `IDetectionRule<T>` but only Python implements it. `PycryptoCryptoHash` is not registered, and the captured value never becomes `SHA-512/256`. `docs/DETECTION_RULE_STRUCTURE.md` is not updated.

The same keyword handling is also missing in `extractArgumentFromMethodCaller` and `PythonSemantic.resolveValues` — that one probably matters more in real code, and needs no DSL change.

engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
san-zrl: Removed the if.

engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
san-zrl: Resolved. Execution order is
Step | Line | What happens
-- | -- | --
1 | 490–504 | Pre-flight loop: every required named parameter is checked for presence. If any is absent → return before anything is written to the store.
2 | 507–510 | MethodDetection emitted — only reached if step 1 passed.
3 | 524–533 | Extraction loop: the else { return; } at line 532 is now a defensive dead branch — the pre-flight already ensures every required parameter exists.

The stale comment "rule does not fire" was the misleading part, because by that point the finding had already been emitted (step 2 runs before step 3). The updated comment now accurately reflects that this path is unreachable in normal operation.

engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
san-zrl: #### PythonDetectionEngine.java
Inside the existing if (hasNamedParams) guard block — after the unknown-keyword check but before MethodDetection is emitted — a pre-flight loop now scans all declared parameters (line 487):

```java
// Reject if any required named parameter is absent from the call site.
// This must run before MethodDetection is emitted so that a call that is missing a
// required argument produces no finding at all (not just a missing child detection).
int preCheckIndex = 0;
for (Parameter<Tree> p : detectionRule.parameters()) {
    if (p.getKeywordName().isPresent() && !p.isKeywordOptional()) {
        Optional<Argument> present =
                findArgumentByKeyword(p.getKeywordName().get(), preCheckIndex, arguments);
        if (present.isEmpty()) {
            return;   // required arg absent → no finding emitted
        }
    }
    preCheckIndex++;
}
```

The loop reuses findArgumentByKeyword (the same resolver used later in the extraction loop) so keyword lookup and positional-fallback semantics are identical. Because preCheckIndex advances for every parameter — not just required ones — the positional fallback slot stays correct relative to the rule's declared parameter order.

####  SHA512NewNegativeRequiredNamedParamAbsentTest.py
A plain SHA512.new() call with no # Noncompliant marker — verifyNoIssue asserts zero findings.

#### PycryptoCryptoHashSHA512RequiredParamTest.java
A self-contained test class that wires a rule variant where truncate is optional=false. testNegativeRequiredNamedParamAbsent calls verifyNoIssue and its asserts() override throws if it is ever reached, making the intent explicit.

engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
san-zrl: Fixed the main example.

#### PythonDetectionEngine.java:
- Line 579 (named-parameter path): replaced checkCurrentIndexState(positionalIndex, …) with checkSymbolTraceState(isInvocation, traceSymbol, expressionTree). The arity guard is no longer invoked because findArgumentByKeyword already guarantees the argument exists before this point is reached.
- Lines 677–693 (checkCurrentIndexState): the symbol-trace body was extracted into the new private method checkSymbolTraceState; checkCurrentIndexState now delegates to it after its arity guard, so the positional path is completely unchanged.
- The new checkSymbolTraceState() carries a Javadoc that makes the separation explicit and points back to findArgumentByKeyword as the reason the arity guard is not needed there.

python/src/main/java/com/ibm/plugin/rules/detection/pycrypto/hash/PycryptoCryptoHash.java
san-zrl: This is intentional. The rule is currently only used in python/src/test/java/com/ibm/plugin/rules/detection/pycrypto/hash/PycryptoCryptoHashSHA512Test.java
to demonstrate the new feature. One this PR is accepted add rewritten versions of all existing Pycryptodome rules.

python/src/test/java/com/ibm/plugin/rules/detection/pycrypto/hash/PycryptoCryptoHashSHA512Test.java
san-zrl: #### PycryptoCryptoHashSHA512Test.java
Added a per-instance flag, defaulting to false, set explicitly by each test method before calling verify(). Because JUnit creates a fresh instance per test, there is no shared-state risk between tests.

Each test method sets the flag
- the 6 fixtures that pass truncate="256", set expectTruncateChild = true
- the 4 that don't set it false.

asserts() now has two hard branches instead of a soft if (store != null)

case | assertion
-- | --
expectTruncateChild = true | assertThat(truncateStore).isNotNull() + full value check
expectTruncateChild = false | assertThat(truncateStore).isNull()

python/src/test/java/com/ibm/plugin/rules/detection/pycrypto/hash/PycryptoCryptoHashSHA512Test.java
san-zrl: Engine is fixed. Test is green. It also asserts the detection of truncate.

engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
san-zrl: Turn log into debug and print file path, line and column.

```java
                    LOG.debug(
                            "Named parameter \"{}\" resolved via positional fallback at index {},"
                                    + " but a preceding optional named parameter was not found."
                                    + " The positional argument may belong to a different parameter."
                                    + " File: {}, line: {}, column: {}",
                            keywordName.get(),
                            positionalIndex,
                            detectionStore.getScanContext().getRelativePath(),
                            expressionTree.firstToken().line(),
                            expressionTree.firstToken().column());
```

engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
san-zrl: Every occurrence of positionalIndex and preCheckIndex removed. Use parameter.getIndex() instead.

engine/src/main/java/com/ibm/engine/language/python/PythonDetectionEngine.java
san-zrl: PythonDetectionEngine.java:

```java
            // Reject if any keyword argument does not match any rule parameter — neither by
            // keyword name nor by positional index. The index check handles mixed rules where a
            // leading withMethodParameter slot is passed by name at the call site, which Python
            // allows.
            List<Parameter<Tree>> params = detectionRule.parameters();
            boolean hasUnknownKeyword =
                    IntStream.range(0, arguments.size())
                            .anyMatch(
                                    i -> {
                                        Argument a = arguments.get(i);
                                        if (!(a instanceof RegularArgument ra)
                                                || ra.keywordArgument() == null) {
                                            return false;
                                        }
                                        String name = ra.keywordArgument().name();
                                        return params.stream()
                                                .noneMatch(
                                                        p ->
                                                                p.getKeywordName()
                                                                                .map(name::equals)
                                                                                .orElse(false)
                                                                        || (p.getKeywordName()
                                                                                        .isEmpty()
                                                                                && p.getIndex()
                                                                                        == i));
                                    });
            if (hasUnknownKeyword) {
                return;
            }
```

A keyword argument at index i with name `n` is "unknown" if no rule parameter satisfies:
  (a) p.getKeywordName() == n                      —> named param whose declared name matches
  (b) p.getKeywordName().isEmpty() && p.getIndex() == i   —> unnamed param at the same slot

Also added PycryptoCryptoHashSHA512MixedRuleTest.java with a hypothetical rule

```java
                    .withMethodParameter(ANY)
                    .withNamedMethodParameter("truncate", "str", /* optional */ true)
```

and corresponding test files SHA512MixedRuleWithTruncateTest.py (detecting truncate correctly)
```python
from Crypto.Hash import SHA512

# Rule: withMethodParameter(ANY) + withNamedMethodParameter("truncate", "str", optional=true)
# data slot is unnamed/required (index 0), truncate is optional named (index 1).

# --- truncate present → must fire and capture truncate ---

# data positional, truncate positional fallback
h = SHA512.new(b"msg", "256") # Noncompliant {{(MessageDigest) SHA-512}}

# data positional, truncate by keyword name
h = SHA512.new(b"msg", truncate="256") # Noncompliant {{(MessageDigest) SHA-512}}

# data by keyword (unnamed slot — Python allows naming positional params), truncate by keyword
h = SHA512.new(data=b"msg", truncate="256") # Noncompliant {{(MessageDigest) SHA-512}}

# truncate by keyword, data absent — must NOT fire (data is required); no Noncompliant marker

# both by keyword reordered: NOT included — the unnamed slot has no declared keyword name so the
# engine cannot match 'data=' at a non-natural position; that call site is unsupported for mixed rules.
```
and SHA512MixedRuleNoTruncateTest.java

```python
from Crypto.Hash import SHA512

# Rule: withMethodParameter(ANY) + withNamedMethodParameter("truncate", "str", optional=true)
# data slot is unnamed/required (index 0), truncate is optional named (index 1).

# --- truncate absent → must fire, no truncate child ---

# data positional only
h = SHA512.new(b"msg") # Noncompliant {{(MessageDigest) SHA-512}}

# data by keyword (unnamed slot), truncate absent
h = SHA512.new(data=b"msg") # Noncompliant {{(MessageDigest) SHA-512}}
```

PR #520, opened 2026-08-21
Csharp first comprehensive rules

fynnth:
Hi all,

This adds the first iteration of comprehensive detection rule coverage for System.security.cryptography (without chacha20poly1305 since this is covered by https://github.com/cbomkit/sonar-cryptography/pull/506). 
There are known gaps:
1. parameters are not detected in a proper form (this will be adressed in a separate pr)
2. sub-namespaces like certificates are not covered (which is one of the most common usages for invoked crypto in this library)
3. There is an issue with the line mapping and code exclusion for #if/else directive statements and this will need to be fixeed in the parser itself not here.
4. certain newer features or certain edgecases can break the parser, which results in skipping some files (but around 98-99% should parse and result in findings)

But what is done correctly:
- All constructors for the crypto in System.security.cryptography should be detected by this pr
- if properties are simply set statically this should also be detected and impacts the finding at constructor level

I have tested this in the following manner:
- I set up a docker container that builds all the testfiles and verifies their validity (all that were compilable on linux, some are only for windows)
- I have run the scanner on all the testfiles and on two open source projects (aspnetcore and bitwarden) where it produced accurate findings, but ran in some of the known gaps with the detailed information.

I think this is a good state to make the pr before reworking the engine for parameters and extending to some additional namespaces.

Best regards,
Fynn


san-zrl (approved):
Hi @fynnth - Thanks for this PR. Looks good to me. I think it's important to complement our csharp lang support with some rules. Good to merge imho.

PR #530, opened 2026-08-31
fix: stop endless recursion in resolveValues on assignment cycles

### LINKED ISSUE #525: Regression in 1.6.1 (compared to 1.6.0): StackOverflowError and excessive logging
Hello,

OS: RHEL 8.10
Java version: 21.0.9 Eclipse Adoptium (Sonar auto-provisioned Java)
Maven version: 3.9.12
Maven plugin: org.sonarsource.scanner.maven:sonar-maven-plugin:5.7.0.6970:sonar
Sonar: Data Center Edition 2026.1.0.119033
Sonar Cryptography plugin version: 1.6.1

I ran into a regression in plugin version 1.6.1: Build fails with StackOverflowError. Plugin version 1.6.0 did not have this behavior.

Unfortunately I do not have a reproducer to share, but this piece of log hopefully shows the issue:
```
[INFO] Using ECJ batch to parse 1057 Main java source files with batch size 500 KB.
[INFO] Starting batch processing.
[INFO] The Java analyzer cannot skip unchanged files in this context. A full analysis is performed for all files.
[INFO] Detected constructor definition has more then one argument to resolve. Redefine the rule to explicitly define the param to resolve
[.. previous line repeats about 7k times ..]
[INFO] 8% analyzed
[INFO] Detected constructor definition has more then one argument to resolve. Redefine the rule to explicitly define the param to resolve
[.. previous line repeats about 12k times ..]
[INFO] 10% analyzed
[.. pattern repeats until ..]
[INFO] Detected constructor definition has more then one argument to resolve. Redefine the rule to explicitly define the param to resolve
[ERROR] A stack overflow error occurred while analyzing file: 'XXX.java'
java.lang.StackOverflowError
     at org.sonar.java.model.expression.IdentifierTreeImpl.symbol(IdentifierTreeImpl.java:85)
     at com.ibm.engine.language.java.JavaDetectionEngine.resolveValues(JavaDetectionEngine.java:320)
     at com.ibm.engine.language.java.JavaDetectionEngine.resolveValues(JavaDetectionEngine.java:337)
     at com.ibm.engine.language.java.JavaDetectionEngine.resolveValues(JavaDetectionEngine.java:353)
     at com.ibm.engine.language.java.JavaDetectionEngine.resolveValues(JavaDetectionEngine.java:337)
     at com.ibm.engine.language.java.JavaDetectionEngine.resolveValues(JavaDetectionEngine.java:353)
     [.. lines 337 and 353 alternate for about 1k lines ..]
```

So there is this new log entry which I do not recall seeing earlier:
```
[INFO] Detected constructor definition has more then one argument to resolve. Redefine the rule to explicitly define the param to resolve
```

Builds that do not fail for stack overflow error can also have this log entry thousands of times, e.g.
```
[INFO] Using ECJ batch to parse 11 Main java source files with batch size 500 KB.
[INFO] Starting batch processing.
[INFO] The Java analyzer cannot skip unchanged files in this context. A full analysis is performed for all files.
[INFO] Detected constructor definition has more then one argument to resolve. Redefine the rule to explicitly define the param to resolve
[.. previous line repeats more than 2k times ..]
[INFO] 100% analyzed
```

n1ckl0sk0rtge:
Fixes #525.

## What was wrong

`JavaDetectionEngine.resolveValues` follows a variable's assignments and its initializer to find its value. It had no cycle check. Code like this made it recurse forever:

```java
private String algorithm = "AES/ECB/PKCS5Padding";
private String copy = algorithm;   // initializer edge: copy -> algorithm

public void swap() {
    algorithm = copy;              // assignment edge: algorithm -> copy
}
```

Resolving `copy` jumps between the assignment branch (line 337) and the initializer branch (line 353) forever. This is exactly the alternating stack trace in #525. The existing `selections.size() > 15` guard never fires, because identifier-to-identifier hops do not grow `selections`.

The bug is old and exists in 1.6.0 too. But 1.6.1 exposed it: since e1fdab3b the engine resolves the arguments of **every** method call (to build detached call records), not only the arguments of matched crypto calls. So a cycle anywhere in the scanned code now crashes the scan. The same change caused the "Detected constructor definition has more then one argument" log flood from the issue.

## The fix

The variable branch of `resolveValues` moves into a new method `resolveVariableValues`. It tracks which variable symbols are being resolved on the current path:

- Before following a variable, its symbol is added to a `resolvingVariables` set.
- A variable that is already on the path is a cycle. It returns an empty list instead of recursing.
- When the variable is done, a `finally` block removes it again.

The release step matters: it cuts only real cycles. Two sibling branches (for example two array elements) may still resolve through the same shared variable. A global "visited" set without release would drop the second value.

## Red-green test

`JcaCipherGetInstanceCyclicAssignmentTest` reproduces the issue:

- **Red:** on `main` the test dies with the exact alternating `StackOverflowError` from #525.
- **Green:** with the fix, the scan finishes and the value still resolves **through** the cycle to `"AES/ECB/PKCS5Padding"`, giving the normal `AES-128-ECB-PKCS5` finding. So cutting the cycle does not lose the detection.

## Not in this PR

- The noisy `LOGGER.info` for constructors with more than one argument. #529 already demotes it to `trace`; that part fits better there.
- The wider question of how much work `buildDetachedCall` does per call. That is a performance topic, not a crash, and deserves its own issue.

Verified with `mvn test -pl engine,java`: 164 tests, 0 failures. Spotless and checkstyle clean.

### DIFF
--- a/engine/src/main/java/com/ibm/engine/language/java/JavaDetectionEngine.java
+++ b/engine/src/main/java/com/ibm/engine/language/java/JavaDetectionEngine.java
@@ -48,10 +48,12 @@ import com.ibm.engine.rule.MethodDetectionRule;
 import com.ibm.engine.rule.Parameter;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -301,7 +303,8 @@ public final class JavaDetectionEngine implements IDetectionEngine<Tree, Symbol>
             @Nonnull Tree expression,
             @Nullable IValueFactory<Tree> valueFactory) {
         if (expression instanceof ExpressionTree expressionTree) {
-            return resolveValues(clazz, expressionTree, valueFactory, new LinkedList<>());
+            return resolveValues(
+                    clazz, expressionTree, valueFactory, new LinkedList<>(), new HashSet<>());
         }
         return Collections.emptyList();
     }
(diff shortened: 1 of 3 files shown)


san-zrl (approved):
LGTM. This also fixes the StackOverflow in the CI pipeline when scanning 
`BlockCipherReorganizer.java` with a simpler call stack:

```
[main] ERROR org.sonar.java.ast.JavaAstScanner - A stack overflow error occurred while analyzing file: 'src/main/java/com/ibm/mapper/reorganizer/rules/BlockCipherReorganizer.java'
java.lang.StackOverflowError
	at com.ibm.engine.language.java.JavaDetectionEngine.resolveValues(JavaDetectionEngine.java:318)
	at com.ibm.engine.language.java.JavaDetectionEngine.resolveValues(JavaDetectionEngine.java:337)
	at com.ibm.engine.language.java.JavaDetectionEngine.resolveValues(JavaDetectionEngine.java:337)
	at com.ibm.engine.language.java.JavaDetectionEngine.resolveValues(JavaDetectionEngine.java:337)
...
```

PR #542, opened 2026-09-11
Fix: Resolve C# alias variable detection for cipher properties

### LINKED ISSUE #511: [BUG] C# engine loses cipher settings when the object is used through a second variable
## Summary

The C# engine links a crypto object to its settings by comparing plain variable names. If the object is used through a second variable, the link breaks and the settings are lost.

In `engine/src/main/java/com/ibm/engine/language/csharp/CSharpDetectionEngine.java`:

- `:420` — `return invocation.getObjectTypeName().equals(sym.getName());`
- `:439` — `return assignedId.equals(sym.getName());`

This is the C# form of #8.

**The fix from #390 cannot be ported here.** That PR walks `Symbol.declaration()` in Java and `Symbol.usages()` in Python. C# has neither. `CSharpSymbol` holds only a name string, and its own javadoc says the ANTLR4 grammar gives no semantic resolution. So C# needs a different solution: track assignments while building the tree in `CSharpTreeConverter`, or build a small symbol table for the file.

## Steps to reproduce

`csharp/src/test/files/rules/detection/dotnet/ProbeCSharpAliasTestFile.cs`:

```csharp
using System.Security.Cryptography;

public class ProbeCSharpAlias
{
    public void TestViaAlias()
    {
        var aes = Aes.Create();
        var alias = aes;
        alias.Mode = CipherMode.CBC;
        alias.KeySize = 256;
    }
}
```

Run it through the C# `TestBase` with `CSharpVerifier`.

## Actual behaviour

```
finding=0
  value=AES
  children=3
  node=AES
```

The three property stores are created, but they carry no detected values. Mode and key size never reach the model. The CBOM entry is a bare `AES`.

## Expected behaviour

The same result as the direct form, which already works. `DotNetAESPropertyTestFile.cs` uses `aes` directly and `DotNetAESPropertyTest` asserts:

```
node.asString() == "AES-256-CBC-PKCS7"
```

With the alias, the expected result is `AES-256-CBC`.

## Severity and frequency

- **Severity: medium.** The data is missing, not wrong. Unlike the Go problem in #510, C# under-reports instead of reporting something false. A cipher shows up without its mode or key length, so the CBOM looks weaker than the code really is.
- **Frequency: medium.** Aliasing is normal in C#: passing an object to a helper, storing it in a field, or `using var x = y;`.

## Acceptance criteria

- [ ] A C# object used through a second variable keeps its property settings
- [ ] The repro file above produces `AES-256-CBC`
- [ ] A regression test is added that fails without the change
- [ ] The existing `csharp` module tests still pass

## Notes

Related work:
- #8 — the original report
- #390 — fixes Java and Python. C# was left out on purpose, and that was the right call, because the approach does not carry over.
- #510 — the Go gap, which is a different problem again (no variable filtering at all)

sachin9058:
When a crypto object is used through an alias variable (e.g., var alias = aes; alias.Mode = CipherMode.CBC;), the engine lost the property settings because it compared variable names directly — alias never matched the tracked symbol aes.
Now the engine tracks alias assignments during tree conversion and resolves them when checking variable invocations.
Changes:
- Add alias map to CSharpBlockTree
- Track var X = Y; assignments in CSharpTreeConverter
- Resolve aliases in CSharpDetectionEngine before matching variable names
Tests: Added DotNetAESAliasTest + ProbeCSharpAliasTestFile.cs. All 38 C# tests pass.

FIXES: #511


fynnth:
Hi @sachin9058 

Nice fix for the direct-alias case, Thank you!
One gap I'd flag before merging, related to the Copilot comment on the earlier revision about "initializer validation": the alias map is only ever written, never invalidated on reassignment. visitAssignment bails out early (children.size() < 2) for a plain alias= ...; statement, so an earlier var alias = first; mapping survives forever, even after alias starts pointing at something else.

Repro:
var first = Aes.Create();
var second = TripleDES.Create();
var alias = first;
alias = second;
alias.Mode = CipherMode.CBC;


Mode=CBC gets attached to first (AES) instead of second (TripleDES), which is a misattribution, not just a miss. That's a regression in kind relative to pre-PR behaviour: before this change the setter was simply dropped for both objects (data loss); now it's silently attached to the wrong one (wrong data)
Suggested fix: handle the plain-reassignment branch in visitAssignment instead of just return null.

It would be great if you could handle that as well, otherwise i will include it my next Major pr and we can merge this for now as it is :)

sachin9058:
Hi @fynnth 

Got it thanks for flagging this. The reassignment case is now handled:
- visitAssignment updates the alias map on alias = second; and adds a CSharpIdentifierTree to the statements list so the engine can pick it up
- CSharpDetectionEngine.run maintains a mutable currentSymbol that gets updated when CSharpIdentifierTree statements are processed, so isInvocationOnVariable always compares against the correct variable
- resolveAlias follows alias chains transitively with cycle protection (var a = b; var b = c; → a → b → c)
- The initializer check only records aliases for plain variable references (not method calls or type names)
- The test file now uses a chain of aliases (first → second → alias) to verify the fix without triggering a second detection
All 39 tests pass. Ready for your review whenever you're free.

fynnth:
HI @sachin9058 ,
I tested the new reassignment handling and it seems to run into a similar problem now:


var first = Aes.Create();
var second = TripleDES.Create();
var alias = first;
alias = second;
alias.Mode = CipherMode.CBC;
Expected: only TripleDES gets Mode=CBC. Actual: both AES and TripleDES get it.

-->
09:31:54.126 [main] DEBUG com.ibm.mapper.utils.Utils -- [  enriched  ] (BlockCipher) AES-CBC
09:31:54.126 [main] DEBUG com.ibm.mapper.utils.Utils -- [  enriched  ]    └─ (BlockSize) 128
09:31:54.126 [main] DEBUG com.ibm.mapper.utils.Utils -- [  enriched  ]    └─ (Oid) 2.16.840.1.101.3.4.1
09:31:54.126 [main] DEBUG com.ibm.mapper.utils.Utils -- [  enriched  ]    └─ (Mode) CBC


Could you have another look at your fix?
Best regards :)
Fynn

